PR #138, opened 2015-01-01
Use country key instead of country value for user-country setting

miquel-cabanas:
Use the country code ($cldr key in $countries array) in the forms when setting the country option and saving it to the database so that it is unaffected by changes to the application language.

### DIFF
--- a/application/modules/mailer/controllers/mailer.php
+++ b/application/modules/mailer/controllers/mailer.php
@@ -94,7 +94,12 @@ class Mailer extends Admin_Controller {
 
     public function send_invoice($invoice_id)
     {
-        if (!$this->mailer_configured) return;
+    	if ($this->input->post('btn_cancel'))
+    	{
+    		redirect('invoices');
+    	}
+    	
+    	if (!$this->mailer_configured) return;
 
         $from             = array($this->input->post('from_email'),
                                   $this->input->post('from_name'));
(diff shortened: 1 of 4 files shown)


Kovah:
Looks good.
@Jadaw1n any objections?

Jadaw1n:
What happens with the existing language setting? Should be converted if possible. (if not possible, mention it bright red in the setup)

miquel-cabanas:
Making the user_country a required field in the user settings form could solve the problem.

In an existing installation the database contains the country-value. On the general settings form the country name is selected based on the language-cldr and correctly displayed in that language, but saving the settings from that page only affects the ip_settings table in the database, and does not touch the ip_users.user_country field. And on the user settings form the country field contains nothings and the user has to set again the country and save it.

To avoid users saving an empty country field, I have added a validation rule to make it a required field. The funny thing is that when attempting to save an empty country field the application complains and then **the country field displays the correct country name** and that it suffices to push again the save button to get the right country-key in the database. Thus, making the user_country a required field in the user settings form seems to solve the problem.

Miquel

miquel-cabanas:
Sorry for the mess, I've learnt the hard way that, if there is already a pull-request in a branch, Github appends new commits to that brach into the existing pull-request. That's why I have reverted the commit, created a new branch called "fix_cancel_button_bug", and made the commit to it, and then a new pull-request.

Kovah:
@miquel-cabanas Why did you closed the pull request?

miquel-cabanas:
Ooops!  still learning git/github :-(  I wanted to update my development branch to the upstream one, thus I renamed it as development-old and then created a copy of upstream/development. I didn't think the pull-request would be gone, I (wrongly) assumed that it would be still around.

I will redo the commit to a feature-specific branch and then will make a new pull-request.

Miquel

PR #144, opened 2015-01-07
Issue/143/use country key for user country setting

miquel-cabanas:
Replaces pull-request #138 "Use country key instead of country value for user-country setting" that was closed by mistake. See discussion on the old pull-request.

### DIFF
--- a/application/modules/settings/views/partial_settings_general.php
+++ b/application/modules/settings/views/partial_settings_general.php
@@ -86,7 +86,7 @@
             <select name="settings[default_country]" class="input-sm form-control">
                 <option></option>
                 <?php foreach ($countries as $cldr => $country) { ?>
-                    <option value="<?php echo $country; ?>" <?php if ($this->mdl_settings->setting('default_country') == $country) { ?>selected="selected"<?php } ?>><?php echo $country ?></option>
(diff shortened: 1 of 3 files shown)


Kovah:
@Jadaw1n Could you take a look at it? :smile:

Jadaw1n:
There are more fields which are based on this setting. I'm updating them also to use the new setting.

miquel-cabanas:
Also we must check the language file before loading it to be sure it contains the `$lang['cldr'] = 'en';` line at the top, and otherwise discard loading it and warn the user to update it.

Jadaw1n:
Good point. But I think this can be generalised, because there are a few new language strings. 
Do you think a warning + link in the setup would suffice?

miquel-cabanas:
I would check the language files during setup but also when running the application, just in case a user copies a bad file afterwards.

Jadaw1n:
At runtime: Would require lang() to somehow notify other functions... Can't really see a clean way here. I'd rather have the language string of the base language (english) displayed. This way the user can still use the application even if a particular string is displayed in english.

jyrkij:
I vote for falling back to base language in case a translation isn't found. But can `cldr` have a fallback?

Jadaw1n:
`cldr` is at this time only used to include a translated list of country names. If `cldr` falls back to `en`, then those are shown in English.

PR #152, opened 2015-01-12
Language specific feature/bug

cz-vilda:
1. When using language file, custom fields table origin were named by
table names form language file instead of real tables names. Repaired,
tested at email templates and at database.
2. When creating new custom field, old scritp remove characters with
diacritics from custom fileds names. New script si changing diacritics
characters to non-diacritics form now. It is better to know what field
is it and reduce database errors due to special characters at column
names or duplicites.
3. New helper "mytexthelper" with functionality to change characters
with diacritics to non-diacritics characters.


Kovah:
@Jadaw1n Could you take a look? Seems good to me.

Jadaw1n:
- Changed naming
- Restored preg_replace check, because things like ~ were not filtered out.

PR #155, opened 2015-01-16
Fix issues with missing timezone

### LINKED ISSUE #32: System settings fail silently if timezone is not properly declared
System settings fail silently if timezone is not properly declared.

The code in question is https://github.com/InvoicePlane/InvoicePlane/blob/master/application/modules/settings/controllers/settings.php#L140

This can be frustrating for users installing the software on some hosting providers.

lommes:
fixes #32 and #154

During setup a warning is displayed if there is no date.timezone. It is still possible to continue setup without changing the settings

The hook is executed before the application controllers and sets date_default_timezone_set to date_default_timezone_get which returns UTC if there is no timezone set in ini, via ini_set or in server configuration (available to php). With ENVIRONMENT set to production a warning still appears in log.

### DIFF
--- a/application/config/config.php
+++ b/application/config/config.php
@@ -91,7 +91,7 @@ $config['charset'] = 'UTF-8';
 | setting this variable to TRUE (boolean).  See the user guide for details.
 |
 */
-$config['enable_hooks'] = FALSE;
+$config['enable_hooks'] = TRUE;
 
 
 /*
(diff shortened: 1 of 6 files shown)


Kovah:
Could you Please re-submit this pull request for the `hotfix-1.1.2` branch? As this is like a bug I would like to include it into this branch for v1.1.2 and not 1.2.0.

Kovah:
Solved with #156

PR #157, opened 2015-01-17
Datepicker - lang specific

cz-vilda:
Use lang specific datepicker labels and day order. Needs to update language file, to config first day of week.

### DIFF
--- a/application/language/english/ip_lang.php
+++ b/application/language/english/ip_lang.php
@@ -2,7 +2,7 @@
 
 $lang['cldr'] = 'en';
 //Day of the week start. 0 (Sunday) to 6 (Saturday)
-$lang['week_start'] = 1;
+$lang['week_start'] = 0;
 
 $lang['account_information'] = 'Account Information';
 $lang['active_client'] = 'Active';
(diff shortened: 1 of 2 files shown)


application/language/english/ip_lang.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
 $lang['cldr'] = 'en';
+//Day of the week start. 0 (Sunday) to 6 (Saturday)
+$lang['week_start'] = 0;
```
Jadaw1n: Is this really language specific? I think it's more a religious thing, which means this should be placed in the settings.

cz-vilda: Yes, you have true. I put it here because:
1. Translators 100% knows, how to set right default for specific language/nation.
2. Me as user, I expect, that after changing language, every nation/language specific features are set automatically. (It could be at settings too with override capability).

So I think at future releases currency, position of currency, decimal point... could be placed here too. Of course with ability to edit it by users. So, if no setting is at config (at database) default form language file is used. As you have it now at any OS.

Jadaw1n: Makes sense, thanks.

PR #173, opened 2015-02-08
Products and products families

egrueda:
Added two modules for products and families management


Kovah:
Everything works perfectly and I think it would be awesome to replace the item lookup with products. So this is on the todo list for v1.2.0.
Thanks for your contribution!

PR #175, opened 2015-02-09
Products lookup

egrueda:
1. replace the item lookup with the products for quotes and invoices
2. add a purchase price to the product details for basic expenses
4. add a db query to move all items from the lookup to the products table


Kovah:
I'll take a closer look at it. :)

PR #179, opened 2015-02-20
Added new functionality: Reports/Sales by Date

domatix:
I have made a new funcionality for Reports: Sales by Date. Now, you can generate a report that shows your payments filtered by two dates and you can filter by minimal and maximum value and show prices with or without taxes too.


Kovah:
Sorry but why do you delete content from the development files and why do you replace the language file with a new one that is based on the old file structure?

Please submit additions to the current state only and do not delete anything!

PR #184, opened 2015-02-24
ADD Reports/Sales by Date functionality

domatix:
Added new functionality for Reports: Sales by Date. Now, you can select two dates in a calendar and filter your reports by them. You can select minimal and maximum quantity and values with or withous taxes too.


Kovah:
Looks good to me. :+1:

PR #185, opened 2015-02-24
This year, last year - dashboard bug

cz-vilda:
This year is all year not just this month, and last year is last year not only same month last year ;)


Kovah:
Could you re-submit the request to the `hotfix-1.2.1` branch?

PR #190, opened 2015-03-10
IP-196 Done

domatix:
You need create the custom field. You should put in Label = "Pago" and in Table = "Invoice".

### DIFF
--- a/application/modules/invoices/views/view.php
+++ b/application/modules/invoices/views/view.php
@@ -43,7 +43,7 @@
                     invoice_status_id: $('#invoice_status_id').val(),
                     items: JSON.stringify(items),
                     invoice_terms: $('#invoice_terms').val(),
-                    custom: $('input[name^=custom]').serializeArray()
+                    custom: $('input[name^=custom],select[name^=custom]').serializeArray()
                 },
                 function(data) {
                     var response = JSON.parse(data);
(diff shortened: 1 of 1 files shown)


application/modules/invoices/views/view.php
```diff
@@ -288,14 +288,20 @@ class="form-control input-sm"
...
 
-            <?php foreach ($custom_fields as $custom_field) { ?>
-                <label><?php echo $custom_field->custom_field_label; ?></label>
-                <input type="text" class="form-control"
-                       name="custom[<?php echo $custom_field->custom_field_column; ?>]"
-                       id="<?php echo $custom_field->custom_field_column; ?>"
-                       value="<?php echo form_prep($this->mdl_invoices->form_value('custom[' . $custom_field->custom_field_column . ']')); ?>"
-                    <?php if ($invoice->is_read_only == 1) { echo 'disabled="disabled"';} ?>>
-            <?php } ?>
+             <?php if($custom_field->custom_field_label=="Pago"){?>
```
Kovah: Did you used a custom field to store the payment method in there?

Kovah:
Needs a complete rework... :/

cz-vilda:
I understand this commit well.

You have to write payment method directly to invoice, at Czech republic. Its told by law. For this, I use custom filed "payment method" (at Invoice table) and I change invoice template to use it.

At my point of view is not necessary to change whole system, but custom fields is ready to be used for this. I think, that make possibility custom fields "mandatory"/"not mandatory" (just as option at custom fields settings) and not just text fileds (but also selects) is right way to do it.

But its about discus...

(I have no time to do it now, but after i finish ToDo lists, I do it).

Kovah:
I would rather add a new field to the invoice table than force the user to create a custom field with a specific name to display the the payment method on the invoice.

PR #192, opened 2015-03-13
better check for headers on import

hokascha:
Import now checks if needed headers exist - no matter in what order


Kovah:
Could you please re-submit this pull request but to the development branch?

PR #198, opened 2015-04-06
€ symbol, quarter (small correction)

cz-vilda:
Not need to add € symbol when using format_currency function. 

Quarter could be written with space or slash (but no hyphen). See http://financial-dictionary.thefreedictionary.com/quarter. I prefer slash, because of it menas "of" (like 1st quarter of year...) more than space.


Kovah:
Thanks!! :+1:

PR #199, opened 2015-04-10
IP-232: Add sum to payment history

malexmave:
Adding a sum row to the payment history report. Currently looks like this:

sum

Consider this a WIP, feedback welcome.


Kovah:
Does this also work correctly with decimal amounts?

malexmave:
Yup:
screenshot from 2015-04-11 15 50 08

Kovah:
Perfect. Will be integrated in one of the next versions as I already closed v1.3.0

PR #203, opened 2015-04-13
Ip 233 password issues

Pitma:
removed de-and encoding for invoice password based on reported problems!


Kovah:
Ist schon im hotfix branch: https://github.com/InvoicePlane/InvoicePlane/commits/v1.3.1
Kümmer mich morgen um das release.

PR #206, opened 2015-04-20
Payment method at invoice templates and space as separator

cz-vilda:
Added ability to use new invoice item "payment method" at invoice templates and pull example template "default-payment" using it.

Nonbreaking space as thousand separator and as between amount and currency symbol (option "currency symbol placement").


application/views/invoice_templates/pdf/default-paymnent.php
```diff
@@ -0,0 +1,258 @@
+<!DOCTYPE HTML>
```
Kovah: Typo in filename: `paymnent`

cz-vilda: sorry...

PR #209, opened 2015-04-21
Fixes relative link of the 'overdue invoices' dashboard panel.

mzur:
The link was broken if IP was run from a subdirectory of the domain (e.g. http://example.com/ip/).
Now it uses the `anchor` function that properly appends the url prefix to the link.

I hope this is the right branch since I couldn't find a `hotfix` branch for patch releases.


Kovah:
Thanks! :+1:

PR #211, opened 2015-04-23
IP-243 and IP-242 done

CarlosDom:
IP-243. Now, you can set up default notes for new quotes.

IP-242. New structure for invoices and quotes.


Kovah:
Looks good, thanks!

PR #214, opened 2015-04-29
Added new functionality: Default payment mode for invoices

CarlosDom:
Now you can select a default payment mode only for invoices. When you create a new invoice, it will has the default payment mode set up.


Kovah:
Payment method is still not set in the invoice.
I'll take care of this okay.

CarlosDom:
It si only set in new invoices, not in created invoices. You have to add a new field in table ip_settings. It works for me. Please review it. Thanks

PR #218, opened 2015-05-13
Added support for Joomla SSO

keithfjn:
Allows Joomla users to be logged in automatically to InvoicePlane when logged into Joomla. Users existing in Joomla but not in InvoicePlane will be created automatically with a configurable user type.


Kovah:
Please do not submit any pull requests to the master branch. Please use the development branch instead.

PR #221, opened 2015-05-17
Greek translation minor corrections/addition

karantzis:
Installing from v1.3.3 zip and using the Greek translations to produce the PDF we noticed some minor translation issues on the Greek language.


Kovah:
Please correct any wrong translations in the [official translation repository](http://translations.invoiceplane.com/)

PR #222, opened 2015-05-19
Fix little mistake in lang key name in client form

pauluz:
I also enlarge a little a input-box for this client_name


Kovah:
Thanks buddy! :+1:

PR #223, opened 2015-05-19
Update invoice.php

Pitma:
Dissallow attachment delete in emailview if invoice is set to read_only mode.


Kovah:
Already added this today... :D

PR #225, opened 2015-05-23
Corrected Filename for downloading from pdf preview in browser (Commu…

Pitma:
Community request: Bug which displayed wrong filename after downloading the invoice from pdf brower preview


Kovah:
Kannst du den noch mal in den `1.4.1` branch pushen?

PR #228, opened 2015-06-05
FIXED IP-286 - Discount input fields are not disabled on read-only

Kovah:


Kovah:
Well *****.
That was the wrong issue ID. Correct one would be IP-295 (Items can be deleted even if in read-only)

PR #260, opened 2015-07-19
Introducing variables in item description

aalwash:
This is my first pull request on GitHub, so bare with me if I'm making mistakes :)
I checked the code if you have a common code for replace variables, but couldn't find one.
Tried to replace variables like the other modules do.

Supported variables:
- Only {{{nextYear}}} is supported

This could be handy for recurring invoices.
For example: <i>Hosting of example.com until August <b>{{{nextYear}}</b></i>
Every year the recurring invoice will contain a item with a description for the next year.

Future variables could be:
- {{{nextWeek}}}
- {{{nextMonth}}}

### DIFF
--- a/application/helpers/pdf_helper.php
+++ b/application/helpers/pdf_helper.php
@@ -38,7 +38,7 @@ function generate_invoice_pdf($invoice_id, $stream = TRUE, $invoice_template = N
     $data = array(
         'invoice' => $invoice,
         'invoice_tax_rates' => $CI->mdl_invoice_tax_rates->where('invoice_id', $invoice_id)->get()->result(),
-        'items' => $CI->mdl_items->where('invoice_id', $invoice_id)->get()->result(),
+        'items' => $CI->mdl_items->get_items_and_replace_vars($invoice_id),
         'payment_method' => $payment_method,
         'output_type' => 'pdf'
     );
--- a/application/modules/guest/controllers/invoices.php
+++ b/application/modules/guest/controllers/invoices.php
@@ -74,7 +74,7 @@ class Invoices extends Guest_Controller
         $this->layout->set(
             array(
                 'invoice' => $invoice,
-                'items' => $this->mdl_items->where('invoice_id', $invoice_id)->get()->result(),
+                'items' => $this->mdl_items->get_items_and_replace_vars($invoice_id),
                 'invoice_tax_rates' => $this->mdl_invoice_tax_rates->where('invoice_id', $invoice_id)->get()->result(),
                 'invoice_id' => $invoice_id
             )
--- a/application/modules/guest/controllers/view.php
+++ b/application/modules/guest/controllers/view.php
@@ -40,7 +40,7 @@ class View extends Base_Controller
 
             $data = array(
                 'invoice' => $invoice,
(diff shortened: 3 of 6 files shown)


Kovah:
Hi there! Thanks for the pull request. Unfortunately it's the wrong branch.
Please check out the current development branch, add your changes there and pull again with a reference to this issue: [IP-318](https://development.invoiceplane.com/browse/IP-318) (add IP-318 directly to the title of your commit please).
If you have any question feel free to join [our chat](https://gitter.im/InvoicePlane/InvoicePlane).

marneu:
@aalwash: There are also thoughs on the theme here https://community.invoiceplane.com/t/make-recurring-invoices-more-flexible-with-vars-like-month/1261
btw.: your code looks better...have you thought about recurring copy_invoice

aalwash:
@Kovah I'm sorry, my fault. I'm going to modify the code a bit, combine it with marneu's idea (it's more flexibel)
Whenever I'm done, I will do a pull request on the development branch.

@marneu Your code is more flexibel, since you can build more "complex" date variables for different cases.
So maybe I can make a combination of your code and mine.

What do you mean with recurring copy_invoice? As far as I could see, it copies the raw data, so the description will stay as it is.
When generating the PDF (or invoice guest view), the data will be converted from variables to dynamic dates. Correct me if I'm wrong

marneu:
@aalwash: I like your idea of combining ;)

Within recurring invoice, I need (and was my reason) a new text in each occuring invoice stating the new/next billing period. Thus said it needs to be a variable in my case. You may think the first invoice for a customer is an invoice template per customer.

As far as I understood now, your idea is you leave the variable as it is unless it is printed (created pdf), am I right?

aalwash:
Yes, you're right. It will only change the variable to a date when it's printed or shown in the guest-view.

About your recurring invoice, everytime a new PDF is generated and emailed, the date will change.
For example:
1-jan-2015 you add an recurring monthly invoice with {{{Month+1}}}

On:
- 1-jan-2015, the variable will become "February", PDF is generated and emailed to the customer
- 1-feb-2015, the variable will become "March", PDF is generated and emailed to the customer
- And so on

By the way, this brings me to a new problem.
I shouldn't use the generation time as a base, because it will change by time if you request for a PDF.
Based on my previous example, if you generated the PDF of 1-feb-2015, it will tell you "March".
But if you try to generated the same invoice on 1-march-2015 (let's say you want to download a PDF for your own bookkeeping) it will tell change the variable to "April".
So I should use the invoice date as base, that way invoices in the paste can't change if you re-generate them.

marneu:
@aalwash: You are right with the problem, at the beginning I had the same idea. This lead me to the "copy_invoice" function. Cause, after executing this routine, the date is fix within the db - for view or further prints/pdf execution.

At this time:
 - I use the variables within products as plain text and manually change the values, if needed.
 - When I do a recurring one (which is normal in our case), I send the first one with real text.
 - After that I change to recurring status and replace the text with the variables again.
   - For clients, you can only use the archieved pdf's from now on in these cases.
 - next time, the copy_invoice function does the job for future invoices.

This is not elegant, but it does the trick at the moment - until your solution ist done ;)
I'd like to stay informed on :dart:

PR #266, opened 2015-07-22
remove database.php from git, create it during setup

nerdoc:
This patch should be an easy solution to keep database.php out of the git repo. Upgrades can safely overwrite the whole structure, without overwriting database.php with it's values.
The file is created dynamically during setup.
Additionally, if a DB errer is encountered, a link to the /setup path is added to the error message.


aalwash:
Love it :+1:

nerdoc:
added a better warning message. "file_exists" replaced with "is_file" because of [this](https://secure.php.net/manual/en/function.file-exists.php#97774) - it's a file, not directory.

Kovah:
Did you tested this? I think this may not work because in the setup no database file is generated, just changed..

nerdoc:
I'm no PHP guru. I just read the code and extended it a bit.
I did the setup procedure a few times without errors with the patch. In write_database_config(), the last line is
```
write_file(APPPATH . 'config/database.php', $db_file);
```
which uses write_file  from CodeIgniter (/system/helpers/file_helper.php). The standard $mode which is used there is FOPEN_WRITE_CREATE_DESTRUCTIVE.

So it's not just changed. I adapted the "require" to use database.php **if if exists**, and if not, read from database_empty.php.
It would be even better to change the file name from database_empty.php to database.php.default to make a NOT-php file of it.
If you want I can add this as well.

Kovah:
Sounds good!

PR #267, opened 2015-07-23
IPT-45

aalwash:
Implement page numbering and it's enabled by default on every invoice/quote.
Unfortunately you can't disable it by setting, if that's required, please ask if that's needed.

If you have a custom PDF Footer, then the page number will not work, since I don't where to place the page numbering.
It's posible to add the page numbering in your custom PDF Footer with these 2 variables:
- {PAGENO}  = Current page number
- {nb} = Total number of pages

For more, see mPDF manual: http://mpdf1.com/manual/index.php?tid=109


Kovah:
Issue moved to [IP-320](https://development.invoiceplane.com/browse/IP-320)

lommes:
It should be possible to add the pagenumber to the existing footer by just appending it in a separate div and add clear both in css to make sure that it's in a new line. The footer will grow a little but mpdf should take care of that.

aalwash:
Hi @lommes , like I mentioned in my pull request, it is possible to add the page number to an existing footer.
The only problem is, I don't know where the IP user wants it to be shown.

We can implement your idea, adding an extra div to show the page number below the existing footer and if someone doesn't like it, then they can disable it in the custom.css

aalwash:
@lommes added page numbering to custom footer text https://github.com/InvoicePlane/InvoicePlane/pull/268

PR #268, opened 2015-07-25
IP-320 - Add page numbering for customer footer 

aalwash:
- Modified CSS, added border on top of the div
- Added page numbering to customer footers (easy to disable, by modifying assets/default/css/custom.css)

### DIFF
--- a/assets/default/css/custom.css
+++ b/assets/default/css/custom.css
@@ -6,5 +6,6 @@
    ========================================================================== */
 
 .footer-including-page-number {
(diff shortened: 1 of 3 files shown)


assets/default/css/custom.css
```diff
@@ -6,5 +6,6 @@
    ========================================================================== */
 
 .footer-including-page-number {
```
Kovah: *Never* add anything to the custom.css file. Please move this to the app scss file.

PR #275, opened 2015-07-31
Hide custom fields if empty in clients module (view)

pauluz:


pauluz:
Like in Clients form view or like in invoice or quote view.php

Kovah:
Reference: [IP-327](https://development.invoiceplane.com/browse/IP-327)
I merge this but please create an issue report for bugs and improvements like this. Only very small syntax corrections can be left untracked.

PR #277, opened 2015-08-06
IP-320 - Add page numbering for customer footer

aalwash:
* Modified CSS, added border on top of the div
* Added page numbering to template.css

If you want to disable page numbering, add this in your custom.css
.footer-including-page-number {
     display: none;
}


aalwash:
Going to make a new branch for this fix only
Will make a new PR when I'm done

PR #288, opened 2015-08-26
input 'type' set and submitted values remembered

WebNinjrDesign:
I've added type="email" to show the email keyboard on mobile devices. Also made it remember what you submit so that if you submit a wrong password, you don't have to re-write your email address and vice-versa.
Also you can test this here where I have implemented it: http://pay.webninjrdesign.com/sessions/login


Kovah:
Whoop.. please *never* submit to the master branch. Use the `development` branch instead.

WebNinjrDesign:
*face palm*

PR #291, opened 2015-08-30
Updated 404 Page

WebNinjrDesign:
An updated 404 error page. Cleaner, simpler with no container.
- Now responsive + viewport set.
- Removed ‘404’ from the page title. No need for this, users just need
to know that the page was not found. The number is a side note.
- Added ‘InvoicePlane’ to the title, so that the user knows its apart
of the InvoicePlane environment.
- Updated the ‘code’ element styling to be a bit simpler. I wasn’t sure
if this was actually needed, as I only see ‘p’ tags coming through.

### DIFF
--- a/application/errors/error_404.php
+++ b/application/errors/error_404.php
@@ -1,62 +1,69 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-<head>
-<title>404 Page Not Found</title>
-<style type="text/css">
-
-::selection{ background-color: #E13300; color: white; }
-::moz-selection{ background-color: #E13300; color: white; }
-::webkit-selection{ background-color: #E13300; color: white; }
-
-body {
-	background-color: #fff;
-	margin: 40px;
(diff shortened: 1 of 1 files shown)


WebNinjrDesign:
You can see this in action here: http://pay.webninjrdesign.com/something-that-doesnt-exist

WebNinjrDesign:
Also, not my design / code. Got this one from HTML5 Boilerplate (https://github.com/h5bp/html5-boilerplate). I just think its clean and simple.

aalwash:
I can't check if you did a PR to the master branch or the development branch on my mobile.

Anyway, what I wanted to say, there is one thing I don't like in the change. You hardcover IP in the title, there is a variable that already contains the title Suffix which is a setting that can be changed by the IP user.

aalwash:
Hardcover =hard coded

Kovah:
I think it's okay not to use a dynamic title for the 404 page

PR #297, opened 2015-08-31
General non-functional fixes/updates

WebNinjrDesign:
Lots of small (non functional) tweaks including...
- spacing of variables
- removing trailing commas from arrays
- standardizing single commas in PHP and javascript
- removing type attribute from script tags as taxt/javascript is default in html5
- adding th tags to table head items
- fixed self closing br and hr tags
- updating frontend code to alternative syntax for if and foreach etc if() : endif; foreach() endforeach;
- fix some html indenting

Small overhaul to application/config/mimes.php - Just made it easier to read and maintain - also change svg to just a string as it was a single index array


Kovah:
Merged with rebase in e.g. 91168d5

PR #301, opened 2015-09-04
Added support for decimals for price and amounts

dodomorandi:
This PR should help with some situations, i.e. when there are big amounts or prices, and two decimals could not be sufficient. With this modification the user is able to change the number of decimals for prices and amounts, eventually leading to less approximations.


Kovah:
Did you tested this modification properly? I already worked on such a patch but the system was not able to handle larger amounts because of the used float number type.

dodomorandi:
Mmmmh... I do not have a large dataset to performance tests, but should not depend on the server type? However, if there is any possibility to test the code against large SQL servers, it would be great. Moreover, it does not change the precision unless you change it, therefore smaller databases should only benefit from these changes... Or not?

Kovah:
Not on large SQL servers... it's just about large amounts in PHP.
Please test your settings with amounts like these:

* 123.456.789.123.456,12
* 123.456.789.123.456,12345
* etc

dodomorandi:
Oops, sorry for the misunderstanding. I changed the code a little bit, now the total number of digits is updated, not only the number of decimals. This will be the SQL type of the changed values:
2 decimals -> DECIMAL(10, 2) (as before)
3 decimals -> DECIMAL(11,3)
4 decimals -> DECIMAL(12,4)

Obviously 123.456.789.123.456,12 is truncated to '99.999.999,99', but a higher number of decimal will not influence the integer part of the number.

Kovah:
Sorry, but `DECIMAL(12,4)` is not acceptable. We currently use `DECIMAL(20,2)` and you can't lower it to 12.

dodomorandi:
Sorry to say that, but you are wrong. When you initialize the database, item_quantity and item_price inside ip_invoice_items and ip_quote_items are set as DECIMAL(10, 2). However, you are right about item_subtotal inside ip_invoice_item_amounts and ip_quote_item_amounts, and I changed the code to use a higher precision for these parameters.

If you are expecting a different behaviour, I thing that a couple of "ALERT TABLE" are missing inside 012_1.4.0.sql. If this is the case, I could make these changes and it will avoid different decimal precision for different parameters. What about it?

Kovah:
The current item amounts support up a decimal setting of `20,2` which was altered in the 012_1.4.0.sql file, only the item price is currently set to `10,2`.
Please test you code with minimum decimal settings like `20,2`, `20,3`,...

dodomorandi:
I have to warn you, as I told you before, that item_quantity, the parameter in which is stored the amount of items for invoices and quotes, is set to DECIMAL(10, 2).
However, I tested the original code and my code against high numbers. As expected, both are not working correctly. The reason behind this is not the SQL storage type, but the float approximation inside PHP. In other words, the bcmath is needed to perform better calculation.
Now I need to ask you what to do: I already tested the calculation using the bcmath, and it works. However, this is independent to the other features I am introducing, therefore I do not know if I have to put the bcmath calculation in this PR or not. In any case, I will check for the presence of the module, to avoid useless incompatibility problems.

I also added the decimal support for the "products" section -- I totally forgot about that. Finally, I decided to remove the format changes to item_subtotal, because I think that it is useless to save a subtotal with higher precision. What do you think about it?

P.s.: try to use the original code to generate an invoice with only one article, with 12.345.678,12 for both amount and price, than try to perform the calculation with other things. The result is 152.415.768.242.646,72, but, as you will notice, the invoice table will give you a wrong number. This just to demonstrate you that it is not a matter of SQL types.

Kovah:
I think we both got each other wrong.
I know about the issues with large numbers and that it's a PHP problem. I just wanted to clarify that if you change SQL types, these should be compatible the minimum settings that are currently available (the 20,3) and that's what I meant with the code testing and not about large numbers.

dodomorandi:
Ok, do not worry :) If you look at the patch, it should be clear that I use the current decimal values as the minimum one. Raising the precision only lead to an increase of precision and decimal parameters.

As I wrote before, I added the support for bcmath, which is not required but it is used if available. Honestly, I tested many cases, and everything seems to be fine. However, it is difficult to find every calculation inside the sources, and if you have any ideas about some sort of testing, they are welcome. I already tested the basic behaviour for both invoices and quotes, however I am worried because, if I only missed one sum, the entire calculation will be zero.

There is also another problem: quote_tax_rate_amount need to be of the same size of item_subtotal (or greater, sure), otherwise there could be some problems with the calculation of taxes for large amounts. Now, for both invoices and quotes, this value is stored as DECIMAL(10, 2), but it should be set to DECIMAL(20, 2).

Whenever you have some time, take a look to the changes. Obviously you can see the differences on calculation just activating bcmath inside you php.ini, and everything should work smoothly.

I will wait for you feedback. If you have some ideas about testing, they are welcome. It is not so easy for me to imagine every possible situation.

Kovah:
Could you re-submit this pull request to the new `bcmath` branch. I would like to implement the conditional bcmath calculations everywhere else.

dodomorandi:
Would you prefer the whole PR or only the bcmath part? I will need only a few minutes to separate these modification from the _decimal_ changes.

Kovah:
The bcmath part would be enough. The decimal settings can be left in the development branch.
Thanks for you work! :+1:

dodomorandi:
Thank you, but _we_ have to thank you, you are doing an exceptional work :+1: 
I created a pull request a https://github.com/InvoicePlane/InvoicePlane/pull/304. Now I will fix these conflicts to make decimal changes work again.

PR #318, opened 2015-11-07
IP-360 - Implemented mPDF template file

Rayvw:
When is was designing the invoice PDF template I wanted to use my company's corporate identity to style the invoice. Redesigning this in HTML is a real pain in the ass, while a simple PDF file could fix the problem. This pull request implements the [mPDF template feature](http://mpdf1.com/manual/index.php?tid=349) (not to be confused with the InvoicePlane templating system). To use it, save a template PDF file to `uploads/pdf_base.pdf`. Make sure the body background-color is set to `none`. Normalize.css defaults to white, which while overlay the template.

### DIFF
--- a/application/helpers/mpdf_helper.php
+++ b/application/helpers/mpdf_helper.php
@@ -17,16 +17,15 @@ if (!defined('BASEPATH')) {
  * 
  */
 
-function pdf_create($html, $filename, $stream = true, $password = null, $isInvoice = null, $isGuest = null)
+function pdf_create($html, $filename, $stream = true, $password = null, $isInvoice = null, $isGuest = null, $baseTemplate = null)
 {
     require_once(APPPATH . 'helpers/mpdf/mpdf.php');
 
     $mpdf = new mPDF();
 
-    $base_template = './uploads/pdf_base.pdf';
-    if (file_exists($base_template)) {
+    if (!empty($baseTemplate) && file_exists('./uploads/'.$baseTemplate)) {
         $mpdf->SetImportUse();
-        $mpdf->SetSourceFile($base_template);
+        $mpdf->SetSourceFile('./uploads/'.$baseTemplate);
         $base = $mpdf->ImportPage(1);
         $mpdf->SetPageTemplate($base);
     }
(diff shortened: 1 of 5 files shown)


Rayvw:
I think this feature is unwanted

marneu:
I really would love to see this in an upcoming version and disagree in "unwanted".

Kovah:
Oh, seems that I missed this one.

lckarssen:
I agree with @marneu that this is a 'wanted' feature. I don't even think that the guy who created our corporate style knows how to recreate his template in HTML. 

In fact, I implemented the exact same thing about a month before he did (but it was done too ugly (and based on v1.2.1) to create a pull request). One thing that I found out while trying to implement this was that mpdf (at least the one shipped with InvoicePlane v1.2.1) only accepts PDF v1.4, so conversion may be needed. I used ghostscript to convert: `gs -sDEVICE=pdfwrite -dCompatibilityLevel=1.4 -dNOPAUSE -dQUIET -dBATCH -sOutputFile=output.pdf input.pdf`.

Rayvw:
Nice to see more people needed this. Maybe in the future we can add an upload field just like the logo upload.

lckarssen:
Another thing that I ran into while implementing this for myself was that my PDF template has the address, VAT details, etc. in the proper font, so I had to remove those from the invoiceplane template.

Rayvw:
I did that too. I think you always have to use a custom template. But that already is the case when you want to implement your company's design.

Rayvw:
@Kovah I rebased the pull request on the current development branch.

Kovah:
Could you please add this issue ID to the pull request title: IP-360
Reference: https://development.invoiceplane.com/browse/IP-360

Rayvw:
I created an upload field on the settings page. I named it 'invoice background' to prevent confusion with the view templates. Note that the default templates currently have a white body background, so no PDF background will be visible. The css body background has to be set to 'none' to view it. I don't think that's a big problem, as one generally has to edit the default template to fit in with the background pdf.

PR #322, opened 2015-11-23
IP-353: EmailToSMS feature is missing SQL table modifications, which is causing errors on save

jaskoola:
fix for IP-353


aalwash:
Thanks, the fix is working! :+1:

PR #329, opened 2016-01-09
Update mdl_import.php

profago:
With this small add the client's import process fills all the fields.

if you accept this change you may have to update also the columns reference in https://wiki.invoiceplane.com/en/1.0/system/importing-data adding the relative two columns.


Kovah:
Please change this pull request to use the development branch. Pull requests to the master branch are prohibited.

PR #331, opened 2016-01-12
Update layout.php

martijnb92:
Chrome Frame is retired.


Kovah:
Removal is not needed as IP 2 will get a completely new theme without any Chrome frames etc.

PR #333, opened 2016-01-13
Added suport for having a custom template for each invoice group

dantefff:
This commit adds a extra select in invoice groups for choosing a custom default template. In case there is no template selected everything works as before.

This allows us having different templates for different kinds of documents (proforma/delivery note/...)

There is a small change in database (see application/modules/setup/sql/017_1.4.4-nextRelease.sql). I did'nt know where to put this change so rename this file as you like.


Kovah:
Please change the target branch to `master`

dantefff:
Sure! I'll make another pull request from development to development to ease the merge

PR #334, opened 2016-01-13
IP-109: Added suport for having a custom template for each invoice group

dantefff:
This commit adds a extra select in invoice groups for choosing a custom default template. In case there is no template selected everything works as before.

This allows us having different templates for different kinds of documents (proforma/delivery note/...)

There is a small change in database (see application/modules/setup/sql/017_1.4.4-nextRelease.sql). I didn't know where to put this change so rename this file as you like.

PD: This is a reborn of pull request #333

### DIFF
--- a/application/helpers/pdf_helper.php
+++ b/application/helpers/pdf_helper.php
@@ -27,8 +27,16 @@ function generate_invoice_pdf($invoice_id, $stream = true, $invoice_template = n
     $CI->load->model('mdl_settings');
     $CI->load->model('payment_methods/mdl_payment_methods');
     $CI->load->library('encrypt');
+    $CI->load->model('invoice_groups/mdl_invoice_groups');
+    $CI->load->helper('country');
 
     $invoice = $CI->mdl_invoices->get_by_id($invoice_id);
+
+    if (!$invoice_template) {
+        $invoice_group = $CI->mdl_invoice_groups->get_by_id($invoice->invoice_group_id);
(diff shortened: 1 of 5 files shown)


kintaro1981:
This needs a little cosmetic fix but it's working on 1.5.5
Another thing: the template selected must be prefilled on the "Send mail" form too.

kintaro1981:
I just saw that it use the "template name" instead of the "template ID" in the invoice_group_pdf_template column in ip_invoice_groups

Kovah:
Pull request needs to be merged into a newer branch other than development as this branch is abandoned.

kintaro1981:
Is this been left behind from the latest IP versions?

clockwiseq:
It's possible.  We'll review it and see if we can merge it into the next release (1.5.10)

kintaro1981:
take a look to this too: #614

clockwiseq:
https://github.com/InvoicePlane/InvoicePlane/pull/334

clockwiseq:
Didn't mean to paste that in here.  Anyway, yes, it looks like this is complete, just needs review.  I'll add it to the list for the next release if we can get it reviewed by then.

kintaro1981:
I linked it only to make you aware it exist so you can choose what is better. 👍

nielsdrost7:
https://development.invoiceplane.com/browse/IP-109 updated

PR #339, opened 2016-02-01
Fix invoice copy (and recurring)

fabian:
I think the commit https://github.com/InvoicePlane/InvoicePlane/commit/de08ee0b4bd00e21a5a32e74b14fd1a1213967e0 broke both invoice copying and recurring invoices: When copying invoices in v1.4.5, the items don't get copied and recurring invoices are also missing the items.

This PR removes the first arguments of all calls to `mdl_items->save`, `mdl_invoice_tax_rates->save`, `mdl_quote_items->save` and `mdl_quote_tax_rates->save`.

It also removes the wrong object access to `invoice_id` and `quote_id` in `mdl_items.php` and `mdl_quote_items.php`.

### DIFF
--- a/application/modules/invoices/models/mdl_items.php
+++ b/application/modules/invoices/models/mdl_items.php
@@ -80,9 +80,9 @@ class Mdl_Items extends Response_Model
         $this->mdl_item_amounts->calculate($id);
 
         $this->load->model('invoices/mdl_invoice_amounts');
-        
-        if (isset($db_array->invoice_id)){
-            $this->mdl_invoice_amounts->calculate($db_array->invoice_id);
+
+        if (isset($db_array['invoice_id'])){
+            $this->mdl_invoice_amounts->calculate($db_array['invoice_id']);
         }
 
         return $id;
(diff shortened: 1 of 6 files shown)


application/modules/invoices/models/mdl_items.php
```diff
@@ -80,9 +80,9 @@ public function save($id = NULL, $db_array = NULL)
         $this->mdl_item_amounts->calculate($id);
 
         $this->load->model('invoices/mdl_invoice_amounts');
-        
-        if (isset($db_array->invoice_id)){
-            $this->mdl_invoice_amounts->calculate($db_array->invoice_id);
+
+        if (isset($db_array['invoice_id'])){
+            $this->mdl_invoice_amounts->calculate($db_array['invoice_id']);
```
Kovah: I had to reverse these changes as they break the quote and invoice saving! :warning:

fabian: Ah yes, I see now that this method is called with an object in [application/modules/invoices/controllers/ajax.php](https://github.com/InvoicePlane/InvoicePlane/blob/master/application/modules/invoices/controllers/ajax.php#L48).

Sorry about that! :disappointed:

PR #340, opened 2016-02-02
Load invoice and quote id from item in database

fabian:
Commit https://github.com/InvoicePlane/InvoicePlane/commit/56422bd863b68d220c4f087330280a0e18889135 fixed the saving of invoices and quotes again, however when copying an invoice or a quote the total of the copied invoice or quote would be wrong as `$db_array` would be an array in that case.

This PR fixes the calculation of the total while copying and also works for saving.


fabian:
`master` okay or should I switch to branch `v1.4.5`?

Kovah:
It's already fixed in branch 1.4.6, see #342

fabian:
Perfect :+1:

PR #346, opened 2016-03-05
Allow client custom fields to be used in invoice group identifiers (IP-553)

RyanDwyer:
My preferred invoice format is ABC-123, where ABC is a code specific to
the client and 123 is the incrementing ID for that client. To do this, I
had to make a new invoice group for each client and statically set that
client's code in the identifier. This change allows me to set the client
code using a custom field and use a single invoice group for all my
clients.

An example format string for this is {{{client:Invoice prefix}}} where
"Invoice prefix" is the label of the custom field.

This also works for quotes.

### DIFF
--- a/application/modules/invoice_groups/controllers/invoice_groups.php
+++ b/application/modules/invoice_groups/controllers/invoice_groups.php
@@ -60,6 +60,10 @@ class Invoice_Groups extends Admin_Controller
         $pdf_invoice_templates = $this->mdl_templates->get_invoice_templates('pdf');
         $this->layout->set('pdf_invoice_templates', $pdf_invoice_templates);
 
+        $this->load->model('custom_fields/mdl_custom_fields');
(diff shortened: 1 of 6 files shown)


RyanDwyer:
I just realised that my pull request doesn't fully resolve my use case. I prefer to have an incrementing ID per client (ie. each client starts at 1) which means I have to use separate invoice groups anyway :(. Oh well. Accept it if you think the feature is worth having, but I guess I'll be sticking to an invoice group per client.

Kovah:
This pull request needs to be rebased on top of the current codebase. A lot of things changed and this may not work anymore for the current version.

denysvitali:
@RyanDwyer have you had time to update this PR?

RyanDwyer:
I ended up developing my own invoicing system.

I'm not going to rebase my PR. You can rebase it if you want, or delete it, I don't mind. I think the feature needs more thought anyway... for example, maybe there should be an invoice group setting to specify whether IDs are per client or per group.

PR #355, opened 2016-05-01
IP-301 refix Special character in Client Name breaks invoice/quote

mmilidoni:
When submit the invoice creation frame, the apostrophe on client name did produce a new client with only left split of client name.


Kovah:
@mmilidoni Please rebase your commit on the latest development branch to remove duplicate commits.

Kovah:
@mmilidoni ?

mmilidoni:
I tried to make rebase but there are many conflicts. How can I apply changes without rebase? Maybe, do I can change development branch, directly?

ayman-alkom:
@Kovah I suggest you ti remove the development branch and let the contributors  directly to push to master.
Just a suggestion, because I saw a lot of failed PR because of that.

Kovah:
The pull requests fail because of changes made in the time between the developer pulled, made his changes and pushed back.
A rebase is essential but I would recommend starting with a fresh pull and copy all changes by hand instead of a rebase, this may be easier.

@ayman-alkom No, the master branch is for the latest stable versioned state of the app, not for development. It would not change any rebasing either.

Kovah:
See #375

PR #366, opened 2016-05-22
Table prefix problem

tradesman:
When using "dbprefix" with "swap_pre" to alter the table prefix, I get "Incorrect parameters in the call to native function 'IFNULL'" on the dashboard after creating the user account.

ip_code_sample
ip_error


Kovah:
What is the purpose of this pull request? It seems you didn't changed anything. For issue reports please use the issue tracker: https://development.invoiceplane.com

PR #379, opened 2016-06-22
V1.4.7

joepraveenkumar:
When viewing invoices, I would like to see an option to be able to export the current list of invoices on the screen in PDF, Excel, or CSV format.


Kovah:
⚠️  Please submit feature requests to the [community forums](https://community.invoiceplane.com/) only!

PR #385, opened 2016-07-14
Auto detect base_url without having to specify the subfolder

s2software:
(The URL_SUBFOLDER also produce an error, since URL_SUBFOLDER constant is not defined "out of the box")


Kovah:
I will try to find another solution for the whole URL problem

PR #417, opened 2016-12-09
Really fixes IP-457 (tested against the exploit)

denysvitali:
As per email sent to mail@invoiceplane.com, this fixes the IP-457 bug that allowed any unauthenticated user to change anyone's password.


Kovah:
The token check already exists, the fix is improved in #418
Thanks anyway for the hint! 👍

PR #418, opened 2016-12-10
Improves the fix for IP-457

Kovah:


denysvitali:
I'm pretty sure the problem still exists, since the token can be "" in the POST request.
Therefore the check is made against the database, which has a "" stored if the user never requested a token.
Have you tested it?

Kovah:
@denysvitali It's now fixed in #419 
Tested all possible exploit options, none worked.

denysvitali:
Thank you 👍 
I can confirm that #419 fixes the issue

PR #419, opened 2016-12-11
Addition to #418

Kovah:
Another fix for IP-457


denysvitali:
I can confirm that this fixes the IP-457 bug.
Thanks!

PR #423, opened 2016-12-23
Created "customer_files" folder in uploads folder

iannato:
I created "customer_files" folder in /uploads folder because of an error i got when installing invoice plane. 

invoicsetuperror

creating the customer_files folder in the /uploads folder fixes this issue


Kovah:
Already done in d5a58fe

PR #427, opened 2017-01-26
Adding some info to invoice's filename

rospus:
In order to save the invoices with a more communicative filename I added also invoice date and customer's name.
Original: Invoice_25.pdf
Modified: Invoice_25_2016-12-31_Filer Acme Ltd.pdf

### DIFF
--- a/application/helpers/pdf_helper.php
+++ b/application/helpers/pdf_helper.php
@@ -75,7 +75,7 @@ function generate_invoice_pdf($invoice_id, $stream = true, $invoice_template = n
     $html = $CI->load->view('invoice_templates/pdf/' . $invoice_template, $data, true);
 
     $CI->load->helper('mpdf');
-    return pdf_create($html, trans('invoice') . '_' . str_replace(array('\\', '/'), '_', $invoice->invoice_number),
+    return pdf_create($html, lang('invoice') . '_' . str_replace(array('\\', '/'), '_', $invoice->invoice_number . '_' . $invoice->invoice_date_created . '_' . $invoice->client_name),
         $stream, $invoice->invoice_password, true, $isGuest, $include_zugferd, $associatedFiles);
 }
 


application/helpers/pdf_helper.php
```diff
@@ -75,7 +75,7 @@ function generate_invoice_pdf($invoice_id, $stream = true, $invoice_template = n
     $html = $CI->load->view('invoice_templates/pdf/' . $invoice_template, $data, true);
 
     $CI->load->helper('mpdf');
-    return pdf_create($html, trans('invoice') . '_' . str_replace(array('\\', '/'), '_', $invoice->invoice_number),
+    return pdf_create($html, lang('invoice') . '_' . str_replace(array('\\', '/'), '_', $invoice->invoice_number . '_' . $invoice->invoice_date_created . '_' . $invoice->client_name),
```
tooster: Why change the trans() to lang()? I believe trans is needed for localisation.
The rest looks good.

aalwash:
Why would you do this? For you it could be intertesting, but other people are OK with only the invoice + number

I prefer making this variable and a setting, so people can make their own filenames combinations

rospus:
2017-01-26 19:37 GMT+01:00 Max De Vos <notifications@github.com>:

> *@tooster* requested changes on this pull request.
> ------------------------------
>
> In application/helpers/pdf_helper.php
> <https://github.com/InvoicePlane/InvoicePlane/pull/427#pullrequestreview-18690813>
> :
>
> > @@ -75,7 +75,7 @@ function generate_invoice_pdf($invoice_id, $stream = true, $invoice_template = n
>      $html = $CI->load->view('invoice_templates/pdf/' . $invoice_template, $data, true);
>
>      $CI->load->helper('mpdf');
> -    return pdf_create($html, trans('invoice') . '_' . str_replace(array('\\', '/'), '_', $invoice->invoice_number),
> +    return pdf_create($html, lang('invoice') . '_' . str_replace(array('\\', '/'), '_', $invoice->invoice_number . '_' . $invoice->invoice_date_created . '_' . $invoice->client_name),
>
> Why change the trans() to lang()? I believe trans is needed for
> localisation.
> The rest looks good.
>
I'm using the italian version of InvoicePlane that has some subtle code
differences, sorry for realising. I'll fix right away, thanks for the
advice.

application/helpers/pdf_helper.php
rospus: You're right. A leftover from italian IP code changes.

rospus:
@aalwash - I did it for myself because in Italy we have such an oppressive fiscality that an accountant is mandatory. With hundreds (I'm not kidding...) of duties every year, almost every week I have to look for specific invoices and finding them by just reading a directory is very helpful. I posted in the italian forum and I was asked to submit it here.
I realize that probably is helpful just in countries like mine and agree with you that making a variable and a setting would be the best thing.
About pull request to the master, I apologize as I'm not a real programer and I'm not used to deal with the development comunity.
If you can point me to right procedures next time I'll operate more correctly. :-)

aalwash:
@rospus  no problem, you're trying to contribute.

The contribution guidlines https://invoiceplane.com/development + https://devwiki.invoiceplane.com/contribution (this one is for IP2 I think, but still the general guide lines are the same)

kintaro1981:
@aalwash maybe can be implemented as are "invoice names" are, so the name of the file could be configurable in the invoice group page.

aalwash:
@kintaro1981 I agree, that's what I proposed in my first comment. It should be a variable filename, if this feature is needed by more people

rospus:
I also agree with Ali Alwash about how implementing the filename feature.

Just a small follow-up as I was contacted by several italian users that
asked my help in order to implement the filename's mod. So it's a feature
that is more tied to the (insanely) complex fiscal system that we have in
Italy. Just to give an idea, the yearly fiscal obligations for a small free
lance are >100 that are impossible to fulfill without an accountant (with
its costs).
Having such a filename feature helps us italian subjects to manage invoices
more easily, especially when you must check accountant's job.

Alessandro

2017-04-10 8:56 GMT+02:00 Ali Alwash <notifications@github.com>:

Kovah:
The client name really bothers me. There will be a lot of escaping needed so filenames will not break on different operating systems.

Also, could you please make sure that you use the 1.5.0 branch instead of master? More information are in the [contribution guide](https://github.com/InvoicePlane/InvoicePlane/blob/master/CONTRIBUTING.md#3-contributing-code).

aalwash:
Beside the wrong, my advice would be, make a feature request in the community forum.
If a lot of people need this, maybe a developer has time to fix it.

About @Kovah's worries with the special chars, which I totally agree with. Maybe the feature should talk about sanitizing the client name before using it, Remove special chars, that most(or at least one) of the OS don't support.

Kovah:
Closing this for now. Need a cleaner solution for that, maybe also a way to customize the filenames.

PR #441, opened 2017-02-12
Suggestion: "data-original-title" to "title" attribute

timacdonald:
The script itself moves the `title` to `data-original-title`. That way the `title` attribute can be a fallback when JS is not enabled. I believe this is the "correct" way to use this plugin.

Not a biggie, just a suggestion.


Kovah:
Hi!, could you please put your changes into the `v1.5.0` branch instead of development and submit a new pull request?

PR #446, opened 2017-02-14
CSS improvement: icons on same line as text in rest. menu

timacdonald:
Moved the css to a wider media query so they are on the same line the whole time the mobile menu is active, previously it only occurred at the smaller size.

[Tablet menu without fix](http://pasteboard.co/y7VURAlU5.png)
[Tablet menu with fix](http://pasteboard.co/y7WGmReS4.png)


timacdonald:
Note: this refers to the right hand float menu items only

PR #468, opened 2017-03-13
Custom Fields Improvements, Graphics Improvements, BugFixes

denysvitali:
- Custom Fields don't require any table customization
- Custom fields are now sorted by order number and label name, take a look at the new custom field form for a better explanation
- Custom fields can now be placed in other groups like "Contact Information", "Taxes Information", "Personal Information"
- Sumex fields are now hidden if the option isn't enabled in settings

@Kovah , this PR doesn't have the migration tool yet.
@Anyone who wants to try this: Upgrading with the included sql file will delete some data from the custom fields, use at your own risk: **don't use it on production data if you don't know what you're doing!**


denysvitali:
@Kovah , about the "Needs revision" tag: should I do something?

Kovah:
Yes, see the code review

assets/core/scss/_core.scss
```diff
@@ -305,14 +304,17 @@ textarea {
 */
 
 #main-area {
-  display: table-cell;
```
denysvitali: Why not?
http://caniuse.com/#feat=flexbox

Kovah: I understand that Caniuse says it's supported but we had several problems even with simple flexbox layouts, especially in Safari both for Mac and iOS. That's why we never really use flexbox for critical elements like the base layout.

denysvitali: Sorry to bother you, but when was the last time you had problems due to flexbox?

denysvitali:
👍

PR #470, opened 2017-03-19
add width and height for personal logo

nicolasbettag:


Kovah:
Is this looking good for small logo images? I would assume that they look blurry if they are stretched to fill the complete width

denysvitali:
Maybe we could use a: ˋmax-width: 100%;ˋ instead of a ˋwidth:100%ˋ

nicolasbettag:
Fixed! 💯

PR #483, opened 2017-04-28
forbid direct access to the ipconfig (IP-517)

zeitschlag:


Kovah:
This is already working. It outputs `#No direct script access allowed`
But the dash is needed for the Env helper to work correctly.

zeitschlag:
@Kovah Have you checked your mails?

PR #487, opened 2017-05-01
Solves 404 (IP-506)

martijnb92:
On some installations


martijnb92:
Guests will get an login screen, it might be better to be able to pay without an account.

Kovah:
Thanks for the pull request! 👍

PR #503, opened 2017-06-14
Added more effective search results on client names - IP-544

musapinar:
Ajax type-ahead for client names would'nt return results if you started typing "ohn" or "hn" while searching for a client named "John Doe" for example.
It now does.

https://community.invoiceplane.com/t/topic/4471/3

### DIFF
--- a/application/modules/clients/controllers/Ajax.php
+++ b/application/modules/clients/controllers/Ajax.php
@@ -37,9 +37,9 @@ class Ajax extends Admin_Controller
         $escapedQuery = str_replace("%", "", $escapedQuery);
         $clients = $this->mdl_clients
             ->where('client_active', 1)
-            ->having('client_name LIKE \'' . $escapedQuery . '%\'')
-            ->or_having('client_surname LIKE \'' . $escapedQuery . '%\'')
-            ->or_having('client_fullname LIKE \'' . $escapedQuery . '%\'')
+            ->having('client_name LIKE \'%' . $escapedQuery . '%\'')
+            ->or_having('client_surname LIKE \'%' . $escapedQuery . '%\'')
+            ->or_having('client_fullname LIKE \'%' . $escapedQuery . '%\'')
             ->order_by('client_name')
             ->get()
             ->result();


Kovah:
Could you change the pull request  to the 1.5.4 branch?

denysvitali:
I'm sorry but I don't agree with this PR. I think that searching "in the middle" is not intuitive at all and may bring misleading results.

Kovah:
Could you give an example where this search may come up with misleading results?

denysvitali:
If I have many clients, and I want to assign an invoice to "Denys", I start typing ""De".
With the original version this will only return results like:
- Debora Kauz
- Denys Vitali

But with this change, results containing "de" will appear too:
- Merce*de*s Bauer
- Denys Vitali
- Debora Kauz

This IMHO becomes a problem when you have many clients, and I really doubt one would search for "gle" when he is looking for "Google Inc."  
It will however works if you search for "Goo" or "Inc.",  because the name starts with "Goo" and ends with "Inc."  
  
Another example:
I'm looking for "Leonard", so I type "Le"  
- Robert Mül*le*r ❌
- A*le*x Richter ❌
- *Le*ah Barth
- *Le*on Eichmann
- Angelika Fied*le*r ❌
- Nico*le* Myers
- Abby A*le*xander ❌  
 The ones marked with a red cross are bad results that shouldn't be shown when I'm searching for "Le".

aalwash:
I partly agree with denys.
A win-win solution would be, add an extra checkbox to search any place in the name.

This checkbox is default should be disabled by default (to keep the current situation as it is)
And the check box value should either be saved in the settings table or in your local browser storage. So you don't have to check it every time (for the ppl that want this as default behavior)

Kovah:
Added the additional task to the issue here: https://development.invoiceplane.com/browse/IP-544

musapinar:
Done within a select box.
Unsure about the translation and if it deserves its own "tab" or not within settings pages.

kintaro1981:
@denysvitali I agree about the "gle" thing but if I want to search a name like Maps and I have these clients:

- Google Photo Inc.
- Google Maps Inc.
- Google Search Inc.

Actually (without that PR) I can't get any result. Remember that many clients name have their business name and not name & surname. 

Plus I think that many people search with at least three characters and in this way many results will gone away.

Your example with 3 character search:
I'm looking for "Leonard", so I type "Leo"

Robert Müller ❌
Alex Richter ❌
Leah Barth ❌
Leon Eichmann
Angelika Fiedler ❌
Nicole Myers❌
Abby Alexander ❌

kintaro1981:
@musapinar how about beside "Please enter 2 or more characters" in the dropdown of the search?

RamonSmit:
```Plus I think that many people search with at least three characters and in this way many results will gone away.``` 

I have alot of customers that are not business related. It's a 40/60 ratio for me.

musapinar:
@kintaro1981 @Kovah @aalwash @denysvitali @RamonSmit 
Feedback welcome before I pull request: https://www.youtube.com/watch?v=u_Q2GS-LvI0&feature=youtu.be
Need a CSS guru as well, I could not find how to fix the right border of the selectbox to look neat, and that really bothers me...

RamonSmit:
I see, I'm not a CSS guru, but I think I would preferer (for new user) to have some kind of description for the toggle switch, It's kinda unknown what it does if you don't know (unlike us in this thread) what it's used for. 


Other then that, I think I like this concept 👍

denysvitali:
I like the feature, but I'm not sure about how I'll do the implementation. Your concept looks great though, so you have my approval on this 👍 

Let's see what the others think...

@RamonSmit I think the tooltip is enough, but I'm not that good at UX, therefore I may be very wrong

musapinar:
The only problem I could find with the tooltip is that it cant pop on a smartphone to give a hint about what it is.
Since it's a toggle switch that people would use once in a while or once in a lifetime (my guess), I'm not sure if it deserves a new row within the form-group of that modal box, or a bigger button. I prefer simpler UIs.
My concern is about the implementation indeed.

Kovah:
Looks okay for me. Maybe you could put a space between the input and the checkbox because you would have to implement the input group feature from Bootstrap into Select2 and I think it's not worth the effort for this small tweak.

PR #505, opened 2017-06-20
Add support for Multiple Currencies

tudip-technologies:
- Added function 'currencies' in invoice helper file that returns array of currencies with labels and symbols.
- Added a function in numbers helper to format amounts with corresponding currency symbol, thousands separator and decimal point.
- Used this function to format amounts on dashboard, client list and each client templates.
- Added a drop down in invoice form that will list currencies. User can select a currency for an invoice.
- Saved selected currency value in database using AJAX request.
- Modified 'format_currency' method in number helper file. Added extra (optional) parameter that will take currency ID. Currency symbol will be decided by this parameter.
- Passed currency Id parameter to format_currency method in templates.
- Displayed amounts grouped by currencies on dashboard page (in the form $ X + € Y, and so on) for all invoice statuses.
- Added functionality to run report of Payment History filtered with currency.
- Added currency list drop down in Payment History index page to show Payment History report based on selected currency.
- Added logic in status function of clients controller to get all clients having active/inactive status with their amounts balances separated with currencies.
- Added logic in view function of clients controller to get due, paid and total amounts separated with currencies.
- Added queries in client model to get amounts grouped by currencies.

**NOTE:**
- Dollar is considered default currency. User can change currency from currency drop down during invoice creation/updation.
- Added multiple currencies support for Payment History under reports section.
- Amounts with different currencies are shown in this form ($X + €Y and so on)

### DIFF
--- a/application/helpers/invoice_helper.php
+++ b/application/helpers/invoice_helper.php
@@ -104,3 +104,28 @@ function invoice_recMod10($in)
 
     return (10 - $carry) % 10;
 }
+
+function currencies()
+{
+    return array(
+        '1' => array(
+            'label' => 'Dollar',
+            'symbol'=> '$'
+        ),
+        '2' => array(
+            'label' => 'Euro',
+            'symbol'=> '€'
+        ),
+        '3' => array(
+            'label' => 'Rupee',
+            'symbol'=> '₹ '
+        )
+    );
+}
+
+function get_currency_symbol($currency_id)
+{
+    $currencies = currencies();
+    return($currencies["$currency_id"]['symbol']);
+}
+
--- a/application/modules/clients/controllers/Clients.php
+++ b/application/modules/clients/controllers/Clients.php
@@ -37,13 +37,40 @@ class Clients extends Admin_Controller
      */
     public function status($status = 'active', $page = 0)
     {
-        if (is_numeric(array_search($status, array('active', 'inactive')))) {
-            $function = 'is_' . $status;
-            $this->mdl_clients->$function();
+        $status_result = array();
+        if ($status == 'all')
+            $status_result = [0,1];
+        elseif ($status == 'active')
+            $status_result = [1];
+        else
+            $status_result = [0];
+
+        $client_results = $this->mdl_clients->with_total_balance_with_currency($status_result);
+
+        //Create an array of number of clients omitting duplicated client in client results
+        //and assigned empty string for client invoice balance to process further.
+        $clients = array();
+
+        foreach ($client_results as $client) {
+            $clients[$client->client_id] = clone $client;
+            $clients[$client->client_id]->client_invoice_balance = '';
         }
 
-        $this->mdl_clients->with_total_balance()->paginate(site_url('clients/status/' . $status), $page);
-        $clients = $this->mdl_clients->result();
+        //Fill the array created above with invoice balance concatenated with total amounts
+        //for a client in different currencies.
+        foreach ($client_results as $client){
+            if (!empty($client->invoice_currency)) {
+                $amount = $client->client_invoice_balance;
+                $symbol = get_currency_symbol($client->invoice_currency);
+
+                if ($amount != 0) {
+                    if ($clients[$client->client_id]->client_invoice_balance)
+                        $clients[$client->client_id]->client_invoice_balance = format_string_type_currency($amount, $symbol, $clients[$client->client_id]->client_invoice_balance);
+                    else
+                        $clients[$client->client_id]->client_invoice_balance = format_string_type_currency($amount, $symbol);
+                }
+            }
+        }
 
         $this->layout->set(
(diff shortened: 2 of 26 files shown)


musapinar:
Cool features.
Shouldn't default currency be editable through General Settings though? I don't like how it's enforced:
`class="form-control" value="$" disabled="disabled">`
If your default currency is not USD, you have to select currency everytime. Am I right?

application/helpers/invoice_helper.php
```diff
@@ -104,3 +104,28 @@ function invoice_recMod10($in)
 
     return (10 - $carry) % 10;
 }
+
+function currencies()
```
denysvitali: Why only 3 currencies are available? Plus, "Dollar" is too generic, USD should be used instead

application/modules/clients/controllers/Clients.php
```diff
@@ -37,13 +37,40 @@ public function index()
...
+        //Fill the array created above with invoice balance concatenated with total amounts
+        //for a client in different currencies.
+        foreach ($client_results as $client){
+            if (!empty($client->invoice_currency)) {
+                $amount = $client->client_invoice_balance;
+                $symbol = get_currency_symbol($client->invoice_currency);
+
+                if ($amount != 0) {
+                    if ($clients[$client->client_id]->client_invoice_balance)
+                        $clients[$client->client_id]->client_invoice_balance = format_string_type_currency($amount, $symbol, $clients[$client->client_id]->client_invoice_balance);
```
denysvitali: I personally am not a fan of this implementation. Currency should be added in another field. There are plugins that rely on the value itself (Zugfred + Sumex) and this will almost certainly break compatibility

application/modules/clients/views/partial_client_table.php
```diff
@@ -17,7 +17,7 @@
                 <td><?php echo anchor('clients/view/' . $client->client_id, htmlsc(format_client($client))); ?></td>
                 <td><?php _htmlsc($client->client_email); ?></td>
                 <td><?php _htmlsc($client->client_phone ? $client->client_phone : ($client->client_mobile ? $client->client_mobile : '')); ?></td>
-                <td class="amount"><?php echo format_currency($client->client_invoice_balance); ?></td>
```
denysvitali: Same as before, IMHO format_currency should be kept, and currencies should be dealt in that function

application/modules/invoices/controllers/Ajax.php
```diff
@@ -443,4 +445,28 @@ public function create_credit()
...
+            );
+        } else {
+            $response = array(
+                'success' => 0,
+                'symbol' => '',
+                'current_symbol' => ''
+            );
+        }
+
+        echo json_encode($response, JSON_UNESCAPED_UNICODE|JSON_UNESCAPED_SLASHES);
```
denysvitali: Why not just `json_encode($response);` as the other implementations above?

application/helpers/invoice_helper.php
musapinar: +1 on ISO codes.

Kovah:
> Added function 'currencies' in invoice helper file that returns array of currencies with labels and symbols.

I would rather add a new helper for currencies that uses a library that provides information about the ISO-4217 currency codes. For example we could use https://github.com/moneyphp/iso-currencies

> Added a function in numbers helper to format amounts with corresponding currency symbol, thousands separator and decimal point. Used this function to format amounts on dashboard, client list and each client templates. [...] Modified 'format_currency' method in number helper file. Added extra (optional) parameter that will take currency ID. Currency symbol will be decided by this parameter.

I would like to stay with the current format_currency() method but extend it to check if the client has an own currency set or use the global setting instead. The format_currency() method may take the client ID as another argument.

> Saved selected currency value in database using AJAX request.

Is the ajax call used that also saves the invoice? If not, I would like to include the currency as the ISO 3-Char Code for the invoice.
Therefore, add another field in the ip_invoices table that is called `invoice_currency_code` and fill it with the code.

Also, the general settings need to be extended with a new setting for the default currency.

denysvitali:
In order to improve it, I need to fully understand what the requirements are.  
Correct me if I'm wrong:
- Every invoice should have its own currency, if no currency is specified the default one is used
- There should be a possibility to see in the reports / dashboard the revenues by currency

Is there something I'm missing?

Kovah:
Yeah, correct. It's mostly updating the format_currency function to use a per-invoice or per-client currency.

Currencies should also be set per client so the correct currency is already pre-selected while creating a new invoice. Like it's currently implemented with the language.

Kovah:
@tudip-technologies Any news on that?

Kovah:
I just merge this pull request so everyone may work on the feature.

@denysvitali @musapinar

esaesa:
I hope I can see this in main release

PR #509, opened 2017-06-26
Add simple plain text body to email (IP-549)

cjsewell:
Improve spam scores by including plain text email body


Kovah:
Many thanks!

PR #510, opened 2017-06-27
Make it possible to delete projects

thueske:
Due to a wrong database column it was not possible to delete projects.

This is now fixed.

### DIFF
--- a/application/modules/tasks/models/Mdl_tasks.php
+++ b/application/modules/tasks/models/Mdl_tasks.php
@@ -263,7 +263,7 @@ class Mdl_Tasks extends Response_Model
             ->get();
 
         foreach ($query->result() as $task) {
-            parent::save($task->task_id, array('task_project_id' => null));
+            parent::save($task->task_id, array('project_id' => null));
         }
     }
 }


musapinar:
folders, filenames, method names, etc.., all are named "task", but the correct column name is "project_id"?

thueske:
It seems... 🤓

bildschirmfoto 2017-06-27 um 20 48 08

denysvitali:
Can confirm:
screenshot from 2017-06-27 20-52-51

musapinar:
ok but why isn't it called `task_id` ?

thueske:
But why should deleting a project delete all tasks? Tasks can also be added without having an project. Deleting them makes no sense in my eyes because invoices can use the tasks as items.

Kovah:
Instead of deleting a project I would rather 'archive' them. Archiving a project makes if unavailable for further actions but leaves it as a reference for the tasks.
Like invoices can be filtered by their status projects could be filtered by 'active' and 'archived'.

Kovah:
I will merge this pull request to fix the current `Unknown column 'task_project_id' in 'field list'` error while deleting a project.

See https://community.invoiceplane.com/t/topic/4584

PR #513, opened 2017-07-08
Update payment_information.php

martijnb92:
Layout consistency


musapinar:
what about `class="text-right"` (Bootstrap) ?

PR #516, opened 2017-07-13
fixed broken download url

musapinar:
The download link can be found at url `/guest/payment_information/form/ULm****VxoWR60q` where `ULm****VxoWR60q` is your invoice url key.

On a side note: it seems like such links can be forged/reached by the guest even if you set "online payments" to "disabled".


Kovah:
Thanks for your work on the software! Really appreciate it! :+1:

PR #518, opened 2017-07-15
download link bug fix

musapinar:
- Fixed a bug where attached image files would point to a broken url and would not render (missing trailing slash).
- Changed `window.open` to `location.href` due to UX concerns.

On a side note: `$config['csrf_regenerate']` should be set to `false` within `/application/config/config.php` for the upload form to work properly. Otherwise you will not be able to upload consecutive files (csrf token 403 error).

### DIFF
--- a/application/modules/dashboard/views/index.php
+++ b/application/modules/dashboard/views/index.php
@@ -154,7 +154,7 @@
                                 </td>
                                 <td style="text-align: center;">
                                     <a href="<?php echo site_url('quotes/generate_pdf/' . $quote->quote_id); ?>"
-                                       title="<?php _trans('download_pdf'); ?>" target="_blank">
+                                       title="<?php _trans('download_pdf'); ?>">
                                         <i class="fa fa-file-pdf-o"></i>
                                     </a>
                                 </td>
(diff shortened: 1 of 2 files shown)


Kovah:
Thanks for the fixes

However, $config['csrf_regenerate'] should stay on true as it's a security-critical setting. If something needs to be fixed it's the dropzone script. Need to look into that because I already implemented a proper handling for the token but it seems it's not working as expected.

musapinar:
It would not be a security-critical setting if the csrf token was stored alongside the session rather than in a persistant cookie. Why do we use a cookie for it? My guess is because someone at some point decided to mix the csrf token with a bit of javascript, in order to be able to read the cookie from a js file. Now THAT is a security issue since the app won't tolerate httpOnly cookies.
This is ridiculous at a point where it forces contributors to hardcode the csrf name, like `name="_ip_csrf" value="xxx"` while we do have a config key for it: `name="<?= $this->config->item('csrf_token_name'); >" value="xxx"`.
I'm not familiar with dropzonejs yet. A quick fix right now would be to avoid regenerating the token within the file that handles the uploaded documents.
Sutff like this is bad:
`
        params: {
            '_ip_csrf': Cookies.get('ip_csrf_cookie')
        },
`

musapinar:
I'm not familiar with all those fancy frameworks of nowadays since I've always used my very own MVC. I've checked the core files of CI's security lib and you were right: it articulates around cookies exclusively.
The way they generate the hash is cryptographically ok. So maybe changing cookies to sessions should do the trick as a quickfix (`/vendor/codeigniter/framework/system/core/Security.php`). However, I'm not sure how those frameworks behave if you change core files, and I'm not willing to learn CI. Plus I guess there will be a lot of files to updates too and I do not have the time to make huge commits right now. I'll sure keep an eye on it though. I'll have more free time starting from August.
In the meantime, for future reference, here are some good readings:
https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet
https://github.com/mebjas/CSRF-Protector-PHP/wiki

PR #522, opened 2017-07-19
Add non-html section to generated mail

liedekef:
Many mail analyzers classify a mail as being spam if it only provides a HTML-body. This change adds a text-body too, and normalizes the breaks (feel free to remove the normalizeBreaks-calls if not desired).


Kovah:
This was already added for version 1.5.4 which may be released in the next days.
https://github.com/InvoicePlane/InvoicePlane/blob/v1.5.4/application/modules/mailer/helpers/phpmailer_helper.php#L82

However thanks for the pull request. 🙂

PR #527, opened 2017-08-02
Fixed fully empty item throwing an error

musapinar:
Fixed a scenario where a user would inadvertently add a new line for a new item and could not save the invoice or could not delete the empty item: "The Article field is required."

### DIFF
--- a/application/modules/invoices/controllers/Ajax.php
+++ b/application/modules/invoices/controllers/Ajax.php
@@ -54,7 +54,7 @@ class Ajax extends Admin_Controller
                     }
 
                     $this->mdl_items->save($item_id, $item);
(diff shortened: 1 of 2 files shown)


application/modules/invoices/controllers/Ajax.php
```diff
@@ -54,7 +54,7 @@ public function save()
                     }
 
                     $this->mdl_items->save($item_id, $item);
-                } else {
+                } elseif (empty($item->item_name) AND (!empty($item->item_quantity) OR !empty($item->item_price))) {
```
Kovah: Could you change that to the `&&` and `||` syntax?

PR #537, opened 2017-09-19
Change of name in quotes templates (IP-583)

daniel89fg:
Se cambia el nombre a las plantillas pdf y public_web, para que sean más acordes con los presupuestos. Ya que antes ponian facturas.

------

It is renamed pdf and public_web templates, so that they are more in line with budgets. Since before they put invoices.


daniel89fg:
application/views/quote_templates/pdf/InvoicePlane.php to application/views/quote_templates/pdf/QuotePlane.php

and

application/views/quote_templates/public/InvoicePlane_Web.php to application/views/quote_templates/public/QuotePlane_Web.php

Kovah:
Uhm.. that does not make any sense. The template names are regarding the application name and not generic. The templates _provided by InvoicePlane_.

daniel89fg:
Has no sense?. Because a quotation template, is named invoice ?. That confuses me. When I'm in the general configuration of the application, in the quotes section, in the templates puts InovicePlace, and makes me think that I'm using the invoice template, not the quotation.

Kovah:
You could rename them to InvoicePlane Quotes.php and something, but not QuotesPlane.

daniel89fg:
quotes

Is something like this no longer understandable? Whatever the name you put, works well.

Kovah:
Yes but the name should not be quoteplane. InvoicePlane as the project and software name should stay in the name! Take InvoicePlane Quotes.php or whatever but not quotesplane.

daniel89fg:
Perfect, ok.

PR #538, opened 2017-09-19
Update Mdl_custom_fields.php (IP-574)

patrice-guillemette:
This fixes the 500 error on php5.6


denysvitali:
My bad.  

Btw, @Kovah does it still make sense to support PHP <= 5.6?

Kovah:
I think it still makes sense to at least support 5.6 to a level that the application is still usable. At least for some time.5.6 will get security updates until 1st Jan 2019.

denysvitali:
PHP 5.6 was released on August 24, 2014 (*3 years ago*). The active support ended on January 27, 2017 (8 months ago). 
Security support will last until December 31, 2018, as per [Release Calendar](http://php.net/supported-versions.php)

PR #541, opened 2017-09-21
Creating separate credit invoices for invoices (IP-581)

daniel89fg:
En algunos países (España por ejemplo), las facturas de crédito tiene que estar separadas de las facturas generales. Las facturas de crédito deben tener una numeración diferente a las facturas geenrales.

El sistema actual crea las facturas de crédito como si fuesen facturas generales, usando la misma numeración para ambos. Eso es erroneo.

Por otro lado, una factura de crédito en el pdf debe reflejar a que factura general pertenece. El sistema actual tampoco lo hace.

Esta corrección permite:
- Crear facturas de crédito separadas de las facturas generales.
- Las facturas de crédito pueden tener su propia numeración.
- En los pdf de las facturas de crédito se puede ver a que factura original esta enlazada.

Con esta correción se puede configurar las facturas de crédito por separado de las facturas generales. Tanto sus pdf, plantilla web y los emails.

---

In some countries (Spain for example), credit bills must be separate from general invoices. Credit invoices must be numbered differently than general invoices.

The current system creates credit bills as if they were general invoices, using the same numbering for both. That is wrong.

On the other hand, a credit invoice in the pdf should reflect to which general invoice it belongs. The current system does not.

This fix allows you to:
- Create credit invoices separate from general invoices.
- Credit bills can have their own numbering.
- In the pdf of the credit invoices you can see to which original invoice is linked.

With this fix you can set up credit invoices separately from general invoices. Both your pdf, web template and emails.


aalwash:
Isn't this feature too much of a country specific thing?

denysvitali:
@aalwash It may be, but I don't see a problem with that - as long as an opt-in option it is provided.

We have already included Zugferd XMLs (Deutschland) and Sumex1 / [Forum-Datenaustaush](forum-datenaustausch.ch) (Swiss "standard" for medical invoices). Both can be enabled (opt-in) by the end-user.  
  
If such an option is provided in this PR (haven't checked), I don't see why it shouldn't be included 👍

application/helpers/pdf_helper.php
```diff
@@ -98,8 +103,12 @@ function generate_invoice_pdf($invoice_id, $stream = true, $invoice_template = n
         'show_item_discounts' => $show_item_discounts,
         'custom_fields' => $custom_fields,
     );
-
-    $html = $CI->load->view('invoice_templates/pdf/' . $invoice_template, $data, true);
+    
+    if ($invoice->creditinvoice_parent_id) {
+		$html = $CI->load->view('invoice_credit_templates/pdf/' . $invoice_template, $data, true);
```
denysvitali: Why is the template placed in ˋinvoice_credit_templatesˋ instead of ˋinvoice_templatesˋ?

application/modules/invoices/models/Mdl_templates.php
```diff
@@ -33,6 +33,21 @@ public function get_invoice_templates($type = 'pdf')
 
         return $templates;
     }
+    
+    public function get_invoice_credit_templates($type = 'pdf')
+    {
+        $this->load->helper('directory');
+
+        if ($type == 'pdf') {
+            $templates = directory_map(APPPATH . '/views/invoice_credit_templates/pdf', true);
```
denysvitali: Same as above

application/modules/invoices/models/Mdl_templates.php
```diff
@@ -33,6 +33,21 @@ public function get_invoice_templates($type = 'pdf')
...
+        } elseif ($type == 'public') {
+            $templates = directory_map(APPPATH . '/views/invoice_credit_templates/public', true);
```
denysvitali: Same as above

application/views/invoice_credit_templates/pdf/.gitignore
```diff
@@ -0,0 +1,5 @@
+*
+!InvoicePlane.php
```
denysvitali: ... why?

application/views/invoice_credit_templates/public/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+!InvoicePlane_Web.php
```
denysvitali: ... why?

daniel89fg:
@aalwash @denysvitali There are several countries that work just like Spain. However, this system allows to continue working as it has until now.

aalwash:
Oke, I agree with Denys, as long it's a opt-in, then it should be fine

application/helpers/pdf_helper.php
daniel89fg: Personally I like to separate the general invoices from the credit bills. Each one with a different folder.

daniel89fg:
Complete example:

General configuration
credit01

Individual Credit Billing Group
credit02

Custom email templates
credit03

We mark the email template for credit invoices
credit04

Modal when creating the credit invoice
credit05

PDF credit invoice
credit06

application/helpers/pdf_helper.php
Kovah: I would like to keep the templates in one folder as long as we do not have 40 templates for separate use cases.

daniel89fg:
You could leave all the files in the template in the same folder as the general invoices. There would be no problem, everything is organizational.

daniel89fg:
@kovah no problem, if you see better tell me and I put everything in the folder ppr default.

denysvitali:
Pardon my incompetence: what is a "Credit invoice", and how does it differ from a Credit Note?  
Isn't this feature already available in IP?

daniel89fg:
@denysvitali the option of credit invoices is available in InvoicePlane, but treats them as invoices, all together. This is what separates the numbers, templates and emails.

daniel89fg:
@kovah How do you see this improvement? something to change? can it be added to the project?

Kovah:
I have no time to check out the whole code so is there an option to toggle this separation on/off? Because we already had several complaints because of the invoice number generation and I think separating invoices and credit invoices completely could lead to new problems.

daniel89fg:
There is no option to activate or deactivate. But I insist it can be configured exactly the same as the current original, which does not modify anything. It is optional, by default everything is the same, but if you want you can separate the invoices. I think it's best that you test it. Please note that without this option InvoicePlane does not work for Spain and other countries.

daniel89fg:
@Kovah what reviews do you need?

Kovah:
Review from our side

PR #545, opened 2017-10-08
ERROR FOUND CUSTOM FIELD MODEL LINE 256

meongbego:
1. Error On Mdl_custom_fields.php File On line 256
2. Change of Code Line #256 $field->$field_id_fieldlabel[] To $field->$field_id_fieldlabel
3. Error Cause Client / Forms Page Can not Open
4. After Editing Code In accordance with Point 2 Problems Done And Everything Has Been Tested To Know The Side Effects Of Changing Code And Trial Not Getting Problems


Kovah:
Already solved in #538 and your fix is not correct.

PR #546, opened 2017-10-09
Update Mdl_custom_fields.php

DavidHazeldenUK:
Client Form won't load without adding 0.


aalwash:
Your fix is wrong, because every loop is going to overwrite the previous (index 0) record.

Probably you have an error somewhere else and you thought this fixes it.
Which version of pho do you have?

DavidHazeldenUK:
I have the latest release version, even the latest Github version still throws that error.
Currently I'm running it with a "0", and it's fine.
5.6 currently.
Will look at running it on 7.1+

DavidHazeldenUK:
https://wptavern.com/php-5-6-is-now-the-most-widely-used-php-version

Kovah:
The patch is included in version 1.5.5 which will be released once there are enough patches to justify the work related to a new release.

PR #557, opened 2017-10-29
v1.5.5

Kovah:


application/helpers/echo_helper.php
```diff
@@ -51,3 +51,14 @@ function _trans($line, $id = '', $default = null)
...
+
+/**
+ * Echo for the auto link function with special chars handling
+ *
+ * @param $str
+ * @param string $type
+ * @param bool $popup
+ */
+function _auto_link($str, $type = 'both', $popup = FALSE) {
+    echo auto_link(htmlsc($str), $type, $popup);
```
ProDigySML: This is a fix for the Stored XSS issue that I found in the client email field. This is just provenance for the required identifier.

PR #558, opened 2017-10-31
Login log, made log-path relative

KLelong:
I made the log-path relative to the base of the web-site.
Please refer me to some information if it"s not the right way to do it.


Kovah:
Please continue using #550

PR #563, opened 2017-11-09
[IP-619] 2nd default invoice tax rate

Bayaccounting:
I have updated file to include 2nd default invoice tax rate while creating new invoice.


Bayaccounting:
Is my code is included on v1.6.0 or have i missed any step

Kovah:
The thing is that I'm not that satisfied with this solution. It may make more sense to modify it in a way so the user is able to add as many tax rates as he wants. 
If we implement this I can see the next request coming: "please allow 3 tax rates to be added"

aalwash:
I agree with Kovah, this is a "hard-coded" way and not flexibel for the future if more taxes are required.

Can someone please explain me how this works?
Because IP already has the ability to have different type of taxes (ie, in NL we have high- and low-taxes)

So in your country you apply 2 times a tax?
Or do some products have different tax-rates? If so, IP already supports that...

Kovah:
The idea behind this feature is that you set multiple _default_ tax rates for invoices.
At the moment you can only specify only one but it seems that some users would like to add more than one tax rate to their invoices by default.

Bayaccounting:
In NZ we only require 2 taxes in default. Also I checked ninja invoice & fusion invoice. Both also supports only 2 default invoice taxes. I will see if I can develop code for unlimited default invoice taxes

UnderDogg:
Needs label : 1.6 or 1.5.11 or something. Needs Milestone?

Kovah:
It's not scheduled for a version yet.

PR #565, opened 2017-11-11
[IP-543] - HTML mail templates not working

kalimerre:
> jQuery.html() treats the string as HTML, jQuery.text() treats the content as text.

That's why the template was not displayed but interpreted.


thueske:
Works for me!

trogper:
this issue is still present in 1.5.11

PR #566, opened 2017-11-11
fix download link

kalimerre:
I don't know why it was setup like that as it attends to go in a controller to generate the file and not download with a direct link.
So I've changed it to site_url() to also get index.php if not removed (otherwise it produces a 404 not found)

(Follow this topic : https://community.invoiceplane.com/t/topic/5194/5 it can have a link)


kalimerre:
Added the fix for the default quote template (link was wrong too)

PR #567, opened 2017-11-11
fix real path for windows

kalimerre:
It miss the realpath function for the customer files folder so the condition was not met on Windows 

> C:\wamp64\www\invoiceplane\uploads/customer_files

instead of 

> C:\wamp64\www\invoiceplane\uploads\customer_files

Be careful ! I didn't test it on a Linux server only on my Windows be sure to test it before merge it 
To see if it works just go on a guest invoice URL with an attachment and try to download it.

### DIFF
--- a/application/modules/guest/controllers/Get.php
+++ b/application/modules/guest/controllers/Get.php
@@ -17,10 +17,10 @@ class Get extends Base_Controller
 {
     public function attachment($filename)
     {
-        $path = UPLOADS_FOLDER . 'customer_files/';
+        $path = UPLOADS_FOLDER . 'customer_files'.DIRECTORY_SEPARATOR;
         $filePath = $path . $filename;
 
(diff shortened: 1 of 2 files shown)


aalwash:
The real fix should be here: https://github.com/InvoicePlane/InvoicePlane/blob/3cd04a62bff7e7b8727b9aef2f5b5939d775ab34/index.php#L314

define('UPLOADS_FOLDER', FCPATH . 'uploads/');
define('THEME_FOLDER', FCPATH . 'assets/');

Both constants have /, instead of DIRECTORY_SEPARATOR

kalimerre:
Indeed @aalwash you're right nice catch. 
I've added a commit to fix it in index.php and change Get.php

aalwash:
👍

Kovah:
Ah, saw this too late. Already fixed with 7d5d229ee8832a0750dc93e1c0023b7d7b42db61

PR #569, opened 2017-11-12
[IP-602] & [IP-603] - Invoice/Quote number generated after sent

kalimerre:
https://development.invoiceplane.com/browse/IP-602
The function to generate invoice number was done after the mail is sent that's why the invoice number didn't exist in the mail.

### DIFF
--- a/application/modules/invoices/models/Mdl_invoices.php
+++ b/application/modules/invoices/models/Mdl_invoices.php
@@ -556,11 +556,7 @@ class Mdl_Invoices extends Response_Model
         if (!empty($invoice)) {
             if ($invoice->invoice_status_id == 1) {
                 // Generate new invoice number if applicable
-                if (get_setting('generate_invoice_number_for_draft') == 0) {
-                    $invoice_number = $this->mdl_invoices->get_invoice_number($invoice->invoice_group_id);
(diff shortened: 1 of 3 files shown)


kalimerre:
I've also added IP603 inside as it was also a bug for quote.
(By the way for "quotes" if the option was set to "No" in settings, no number of quote will be assign even after being sent, it should be fixed and in the good order with this fix)

aalwash:
I didn't test the code, but doesn't make any sens to move the invoice number increaser to the Mailer class?
It's a mailer controller, shouldn't it not only care about emailing?

It would make more sense, to add a function in the Mdl_Invoice, which is callable from the Mailer.
So all logic stays in the Mdl_Invoice, something like Mdl_Invoice->generateInvoiceNumberIfApplicable()

kalimerre:
Yep you're right, gonna make the change tonight

kalimerre:
np ! 
(I don't take it as negative comments don't worry I'm happy someone else review my code, I just need some adaptation time to understand correctly the process in this project) 

And thanks for your help and noticing the mistakes :)

kalimerre:
And fixed
thanks @aalwash

PR #573, opened 2017-11-20
[IP-623] - Different footer text between invoice and quote

kalimerre:
https://development.invoiceplane.com/browse/IP-623

New field added in the settings, SQL done, language ok,


application/helpers/mpdf_helper.php
```diff
@@ -75,11 +75,17 @@ function pdf_create(
     }
 
     // Set the footer if voucher is invoice and if set in settings
-    if (!empty($CI->mdl_settings->settings['pdf_invoice_footer'])) {
+    if (!empty($CI->mdl_settings->settings['pdf_invoice_footer']) && $isInvoice) {
```
kalimerre: Little explanation about this, I used $isInvoice because if I use 
`strpos($filename, trans('invoice')) !== false`

The footer will be also included in the report invoice_aging

`trans('invoice_aging')`

(as it seeks for invoice in the title of the file ...) 

If it's really a problem feel free to tell me and I will commit with `trans('invoice')` instead of `$isInvoice`

application/modules/setup/sql/029_1.5.6.sql
```diff
@@ -0,0 +1,3 @@
+#Fix for IP-623
```
Kovah: You can remove that as it will be automatically added if the settings are updated :)

kalimerre:
@Kovah File removed :)

aalwash:
Just a small note, this one needs to be added to the setup file.

With the current code, the quote footers will disappear, since the invoice and quote footer are the same prior to this PR.
After this PR, all quotes will have no-footer, which means, breaking change.

Other then that, it looks good.

PS. Didn't test the code

PR #574, opened 2017-11-26
[IP-362] - Setup the input and modify the language

kalimerre:
Need ToDo

- The setup function to save the admin's email if old option `bcc_mails_to_admin` is set to "1"

I remove the hint in the language as it's not needed anymore, if you really want it then I can re-add it

Not merge before the setup function has been made or users won't have his mails as BCC anymore


application/language/english/ip_lang.php
```diff
@@ -37,8 +37,7 @@
     'back'                                         => 'Back',
     'base_invoice'                                 => 'Base Invoice',
     'bcc'                                          => 'BCC',
-    'bcc_mails_to_admin'                           => 'Send all outgoing emails as BCC to the admin account',
-    'bcc_mails_to_admin_hint'                      => 'The admin account is the account that was created while installing InvoicePlane.',
+    'bcc_mails_to_admin'                           => 'Enter one or multiple email(s) to send outgoing emails as BCC',
```
aalwash: Maybe mentioning, comma seperated :)

application/modules/mailer/helpers/phpmailer_helper.php
```diff
@@ -124,12 +124,13 @@ function phpmail_send(
...
     }
 
-    if (get_setting('bcc_mails_to_admin') == 1) {
+    if (get_setting('bcc_mails_to_admin_email') != "") {
         // Get email address of admin account and push it to the array
-        $CI->load->model('users/mdl_users');
-        $CI->db->where('user_id', 1);
-        $admin = $CI->db->get('ip_users')->row();
-        $mail->addBCC($admin->user_email);
+        $bcc_admin = (strpos(get_setting('bcc_mails_to_admin_email'), ',')) ? explode(',', get_setting('bcc_mails_to_admin_email')) : explode(';', get_setting('bcc_mails_to_admin_email'));
```
aalwash: Please keep it backward compatible. So if bcc_mails_to_admin == 1, then bcc_admin contains the email of the admin user.
If bcc_to_mails_to_admin_email != empty (and the new setting is leading), use that value.

Another thing, either it should support separated by comma or semi-colon, so don't check if comma is in the string.

Plus, you can do explode(',', $var) and still get an array, if even there is no comma in it, so no need of strpos()

kalimerre: For the backward compatible it's in the ToDo, the setup function that has to be made to save the field with the email of the admin user if bcc_mails_to_admin == 1 (function that will be run with the setup) but it has to be done.


Hmm so what do you suggest for the comma/semi-colon problem ? 
To be honest I re-use the same way it was done for the $to variable, $cc, or $bcc (https://github.com/InvoicePlane/InvoicePlane/blob/master/application/modules/mailer/helpers/phpmailer_helper.php#L101) to use the same code at each part and not have any difference between two similar lines but maybe this has to be changed even for `$to` `$cc` `$bcc`

application/language/english/ip_lang.php
kalimerre: Yup thanks, will modify it

application/modules/mailer/helpers/phpmailer_helper.php
aalwash: About the BC, sorry I didn't read your original comment with the PR. You already mentioned that.
With setup, you mean the upgrade setup? Not the "new-setup-wizard"? If yes, then that's fine with me


About the strpos(), in that case, keep it as it is.

kalimerre: No problem mate ;) 
Yeah the upgrade setup like this function (https://github.com/InvoicePlane/InvoicePlane/blob/master/application/modules/setup/models/Mdl_setup.php#L238), 
Like a `public function upgrade_029_1_5_6()` and write the code to check if `bcc_mails_to_admin == 1` if that is the case then insert the `bcc_mails_to_admin_email` field in the DB with the email of the admin user, if not just leave the function. If I don't do it, it's just because I don't want to fucked all my dev env on my computer (delete all, re-install all to run the upgrade function) but yeah the goal is to add a function which will be launched during the upgrade process :)

aalwash: 👍

kalimerre:
Fixed for the translation (comma separated) thanks @aalwash

kalimerre:
The ToDo in my original post
The setup function to save the admin's email if old option bcc_mails_to_admin is set to "1"

UnderDogg:
Needs some work before we can merge. @kalimerre can you work on that Todo please?

PR #575, opened 2017-11-27
[IP-367] - Calculation process and order invoice/item tax/discount

kalimerre:
https://development.invoiceplane.com/browse/IP-367

ToDo : 

- Make some testing again
- Move the CSS to the corresponding file (don't hesitate to commit on this PR)

### DIFF
--- a/application/modules/settings/views/partial_settings_taxes.php
+++ b/application/modules/settings/views/partial_settings_taxes.php
@@ -63,4 +63,112 @@
         </div>
     </div>
 
+    <div class="panel panel-default">
+        <div class="panel-heading">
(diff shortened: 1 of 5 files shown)


application/modules/settings/views/partial_settings_taxes.php
```diff
@@ -63,4 +63,112 @@ class="form-control simple-select">
...
+                });
+            }
+        });
+
+        $( "#sortable1 li, #sortable2 li" ).disableSelection();
+    } );
+</script>
+
+
+<!-- Move it to the CSS file -->
```
kalimerre: This has to be moved in the correct file if the design is ok

PR #578, opened 2017-12-10
[IP-625] Bugfix some wrong columns and index

franciscorangel:
Bugfix some parts of the code, changing all parts to the new column names, this problem started when I have done a update 1.4.x to 1.5.x

Update code to accept composite primary keys and work with invoice custom.
Alter some tables to new column name.

Changed a primary key index.

### DIFF
--- a/application/core/MY_Model.php
+++ b/application/core/MY_Model.php
@@ -220,7 +220,19 @@ class MY_Model extends CI_Model
                 }
             }
 
-            $this->db->where($this->primary_key, $id);
+            if(is_array($id)){
+                $where = [];
+
+                foreach($this->primary_key as $key){
+                    $where[$key] = $id[$key];
+                }
+
+                $this->db->where($where);
+            }
+            else{
+                $this->db->where($this->primary_key, $id);
+            }
+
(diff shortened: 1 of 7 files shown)


Kovah:
@denysvitali Could you check these changes for the custom fields?

denysvitali:
I've read the code, and it looks ok to me. I still have to test it though. I'll take a look at it this week

franciscorangel:
I added other bugfix related, now enable to save custom field with empty value.

Kovah:
@franciscorangel Could you outline what exactly the problem was and which errors occured while migrating from 1.4 to 1.5? Would like to know more about this as custom fields were working perfectly for me and did not encounter any errors.

miquel-cabanas:
I cannot reproduce the problem in v.5.9 and v1.5.10.

Apparently this PR is no longer needed and may conflict with the code in v1.5.9/v1.5.10 (there is already a column named 'client_custom_fieldid') and/or in CI (changes to application/core/MY_Model.php).

I recommend rejecting this PR as no longer needed for v1.5.10.

PR #584, opened 2018-01-28
[IP-510] - Import function not working

kalimerre:
Increase the key length to avoid duplicate and a if statement to allow import (as they have no invoice_group_id they don't need to enter in the if statement)


kalimerre:
https://development.invoiceplane.com/browse/IP-510

PR #590, opened 2018-01-29
[IP-638] Add product filtering

cornelinux:
Add a filter field in the product list view.
You can filter for SKU, product name and product_description.

Closes [IP-638]


cornelinux:
However there is one caveat:
When I run a filtering it works fine. But the product navigation bar (with the "create new product" button) is displayed twice. And I do not know how to remove this.

Kovah:
Will take a look on it after 1.5.6 release

cornelinux:
OK, I understand it being a time issue :-)

PR #596, opened 2018-02-06
Fix for output of Japanese font which was not displayed correctly

kusanaginoturugi:
PDF出力した日本語フォントが正しく表示されていなかったので、設定を追加しました。


Kovah:
Could you please change the title to english and add a quick description?

Kovah:
@kusanaginoturugi

You85:
Translation : "The output Japanese font was not displayed correctly, so the setting was added."

miquel-cabanas:
There is now an issue for this PR:

IP-737 - Fix display of some non-Latin characters in mPDF generated PDF files
https://development.invoiceplane.com/browse/IP-737

PR #597, opened 2018-02-08
[IP-596] Add Email Mailsender

cornelinux:
Closes IP-596


kalimerre:
Hey

Not sure about it but it miss the setup function to add system@ in the DB no ? Because it will appear in the form as empty after the upgrade but if it’s empty it’s replaced by system so it may be confusing for the user. 

Otherwise +1 for me it looks good ! :)

application/modules/sessions/controllers/Sessions.php
```diff
@@ -187,7 +187,11 @@ public function passwordreset($token = null)
                 $email_message = $this->load->view('emails/passwordreset', array(
                     'resetlink' => $email_resetlink
                 ), true);
-                $email_from = 'system@' . preg_replace("/^[\w]{2,6}:\/\/([\w\d\.\-]+).*$/", "$1", base_url());
+
+                $email_from = get_setting('smtp_mail_from');
+                if (emtpy($email_from)) {
```
cornelinux: @kalimerre this is why we check if the database is empty. If it is, it reverts to the old behaviour of using the ``system@`` for backward compatiblity. So that if a user updates, nothing will change in the behaviour.

kalimerre: I saw that and it’s fine for me. The only thing which is not fine is that in the form in the settings the first time it will appears as empty/blank because there is no value in the DB. So user may think it will be empty or something else whereas it should shown system@ as default while going to the form for the first time. 

I don’t know if I’m very clear I’m on my phone so it’s hard to explain ;).

Kovah: Thanks for the hint. The settings label needs to be changed to something more clear, like "From address for system emails" or something similar.
I would not explicitly use "for password reset" emails to continue using the setting for other internal emails like for notifications.

cornelinux: @kalimerre I understand. So maybe something like using ``placeholder=`` would be sensible, so that the user knows, that ``system@`` will be used as default, if he enters nothing.

@Kovah I happily can change this.

kalimerre: Fine for me if it’s good for @Kovah too ;)

application/modules/settings/views/partial_settings_email.php
```diff
@@ -72,6 +72,15 @@ class="form-control simple-select">
                                    class="form-control"
                                    value="<?php echo get_setting('smtp_server_address', '', true); ?>">
                         </div>
+                        
+                        <div class="form-group">
+                            <label for="settings[smtp_mail_from]">
+                                <?php _trans('smtp_mail_from'); ?>
+                            </label>
+                            <input type="text" name="settings[smtp_mail_from]" id="settings[smtp_mail_from]"
```
kalimerre: Should we not check if the input is a mail instead of a text ? We need to be sure that the text entered here is a mail address no ?

cornelinux: You mean 

    <input type=email ...

kalimerre: Yes but on the phone it’s hard to write tags haha ! 

I don’t know if there is other validator in the settings for mail but I guess. Maybe add this field to the fields validated for mails no ?

PR #609, opened 2018-03-04
Bool custom field receive wrong value from other field (IP-640)

Kovah:
Pull Request Checklist

  * [x] Bugfix
  * [ ] New Feature

### DIFF
--- a/application/modules/custom_fields/models/Mdl_custom_fields.php
+++ b/application/modules/custom_fields/models/Mdl_custom_fields.php
@@ -15,7 +15,9 @@ if (!defined('BASEPATH')) exit('No direct script access allowed');
  */
 class Mdl_Custom_Fields extends MY_Model
 {
+
     public $table = 'ip_custom_fields';
+
     public $primary_key = 'ip_custom_fields.custom_field_id';
 
     /**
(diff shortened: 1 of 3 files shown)


application/modules/custom_fields/models/Mdl_custom_fields.php
```diff
@@ -262,7 +264,7 @@ public function get_values_for_fields($custom_field_model, $model_id)
                         $field->$key_serialized .= $custom_value === end($custom_values) ? '' : ', ';
                     }
                 }
-            } else {
+            } elseif ($field->custom_field_type == 'SINGLE-CHOICE') {
```
aalwash: I didn't test this, but is there a reason why you didn't keep an else statement?
Or is that causing the bug to re-occur, even if you added the elseif?

Kovah: I did not write the function itself but as far as I understand `get_values_for_fields()` only handles values for choice fields (multiple or single). The else statement handles all custom field types which leads to wrong values, because there are no values to get for types like text or dates.

aalwash: Oke, sounds fair :)

PR #631, opened 2018-05-20
[IP-711] add responsive view for item lists in quotes/invoices

der-peer:
Pull Request Checklist

  * [ ] Bugfix
  * [ ] New Feature

It's my first try, though having read the guidelines I'm quite sure having done things wrong. Please be patient... it's quite confusing...


der-peer:
I think, the PR is complete so far...
As far as I can tell, this view and the original table view work both as expected and there should nothing be broken.

Three points:
- please disregard lines in custom-pdf.css. It's for my own template, sorry.
- lines from custom.css need to be integrated into style.css... but I don't know how, atm.
- style.css:
`@(max-width: 767px)
#content .row {
    margin-left: -10px;
    margin-right: -10px;
}
`
I don't see, why this is needed... but it breaks nested Bootstrap rows.

Kovah:
Funding will be split 50:50 for development and testing.
50% go to @der-peer 
50% go to a user who tests the feature.

[Test Package](https://github.com/der-peer/InvoicePlane/archive/v1.6.0.zip)

Kovah:
@amitkr936 Please notice that this feature is completed and only needs testing! ⚠️ 
More details about paid rewards: https://community.invoiceplane.com/t/topic/6898/

Kovah:
I think I cannot merge this pull request. The main problem is that there are fixes for calculations and new settings for net/gross amounts which have nothing in common with the responsive view of item tables.
A suitable way to get this into 1.6.0 would be to split this up into two different pull requests. One that only contains changes for the responsive view of the item tables and one for all calculation things.

About the calculation thing: there were critical changes implemented for 1.5.10 (#638) and I am not sure if these changes are compatible with your implementation. However, I would really like to get your net/gross implementation into the app.

PR #654, opened 2018-08-25
[IP-727] Improve handling of decimal and thousand separator symbols

miquel-cabanas:
Pull Request Checklist

  * [X] My code follows the code formatting guidelines.
  * [X] I have an issue ID for this pull request. 
  * [X] I selected the corresponding branch.
  * [X] I have rebased my changes on top of the corresponding branch.
  
  * [X] Bugfix
  * [ ] New Feature

Improve the setting of decimal and thousand separator symbols to make sure users cannot enter a setting that will cause amounts to be wrongly formatted and displayed, e.g. as reported in https://community.invoiceplane.com/t/topic/6742/10

- Move option "Number of items in lists" from "General/Amounts Settings" to "General/General".
- Let users select the number format and set the thousands_separator and decimal_point accordingly.


Kovah:
Awesome. 
Are these the only number formats used? Could imagine that there are other ways to format numbers in some countries. 🤔 
I would at least add one simple format without a thousand separator (`1000000,00` or/and `1000000.00`).

miquel-cabanas:
Oh yes! there are other formats [1] but the US/UK and EU are most common ones [2] and the ISO are the officially endorsed [3]

[1] https://en.wikipedia.org/wiki/Decimal_separator#Other_numeral_systems
[2] https://en.wikipedia.org/wiki/Decimal_separator#Arabic_numerals
[3] https://en.wikipedia.org/wiki/ISO_80000-1

So, I can add the two formats you have mentioned, and if a user comes later with another request, we can add it to the view and controller files.

How should I proceed? do I submit a new PR with the same issue number and the new code changes?

Kovah:
> How should I proceed? do I submit a new PR with the same issue number and the new code changes?
Just make another commit to the `v1.5.10-ip-727` branch.

> should I rewrite the if-then-else block as cases of the switch-case block?
I would entirely remove the switch or if statements and move all available number formats into a configuration file. For displaying options: loop trough options, for saving check if the used config key exist and use these settings.

miquel-cabanas:
I have made the requested changes:
- there is a new configuration file (number_formats.php) containing the list of available number formats
- the configuration is imported into IP as an array
- the dropdown in partial_settings_general is dynamically built from the array
- decimal_point and thousand_symbol are set according to the selected number_format
- only decimal_point, thousand_symbol and number_format are stored in the db (ip_settings table)
- the english language file has been updated.

Note: I have added back to the english language files three definitions (past month, quarter and year) that were removed by mistake in a previous commit.

miquel-cabanas:
Oops! i clicked the wrong button :-(

Kovah:
Full funding will be paid to @miquel-cabanas.

PR #664, opened 2018-10-01
[IP2-70] Support more RDBMS in setup

749:
Pull Request Checklist

  * [ ] Bugfix
  * [ ] New Feature

[Issue](https://development.invoiceplane.com/browse/IP2-70)

image

image


app/Modules/Setup/Controllers/SetupController.php
```diff
@@ -75,7 +75,26 @@ public function prerequisites()
...
+        $drivers = array();
+        foreach (\PDO::getAvailableDrivers() as $driver) {
+            // check if driver is supported by laravel
+            if(config('database.connections.' . $driver) != null) {
+                $drivers[$driver] = $driver;
+            }
+        }
+
+        // prepare the database name if we are using the sqlite driver
+        $current_settings = config('database.connections.mysql');
```
Seb33300: The comment just above refer to sqlite but you are getting the mysql config `database.connections.mysql`?
Is it normal?

Seb33300:
I think SQLite database file should be stored in the `storage` folder and not in the `database` folder.

And maybe we should not let the choice to the user to choose the database filename.
I don't really see any advantage for the user to choose the name of the file, and this make the installation process more complex since the user may not understand the purpose of the field.

app/Modules/Setup/Controllers/SetupController.php
749: The configuration values are replicated across the drivers. I just use the MySQL one as it has all the fields that the user can edit. See the config/database.php file. I will add a comment to explain.

749:
I just used the default path that laravel uses for sqlite, it is easy to change though. Just let me know which version you would like.

749:
Added a message in case no PDO drivers are available:
image

749:
Just noticed: the database backup uses Mysqldump, that will not work sqlite or the other RDBMS.
I suggest we take a look at [https://github.com/spatie/db-dumper](https://github.com/spatie/db-dumper)

PR #703, opened 2020-01-12
Fixing clipboard issues

UnderDogg:
In order to fix some clipboard issues we needed to include clipboardjs.
Then i made sure it was included in the package.json
And in the Gruntfile.js it needs to be compiled to dependencies.js

Finally in /assets/core/js/scripts.js we need to call ClipboardJS instead of Clipboard (current)
* [ ] Bugfix


MartinH0:
Tested it. Works perfectly fine.

- no JS Errors now.
- Click-To-Copy works.
- Variables are working in Email templates again.

PR #704, opened 2020-02-05
V1.6.0

KirkBushman:
I've merged master on branch v1.6.0, and added both before and after-tax calculation based on a switch in settings, as discussed here:

https://community.invoiceplane.com/t/topic/8544/14

To a first comparison, the numbers should be ok for both cases, we need to check the total invoice numbers, that are still wrong. Let's go by steps. I don't want to push too much stuff all at once.
I'll make another commit fixing the total.

In the meantime,
lets' all check the numbers together and make sure we all agree they are correct.


boban-dj:
Hi, 

I have tested and have these issues:

issue 1 
when enabled an item tax and discount in percent or amount, they do not calculate the same

issue 2
when invoice tax is enabled discount percent or amount also not calculate the same

discount is 50% and amount is 50 euro of 100 euro
please see screenshots

Item tax with different values

item_discount-amount_item_tax
item_discount-percent_item_tax

Invoice tax enabled here

invoice_tax_amount_discount
invoice_tax_percent_discount

noplanman:
@KirkBushman Any idea when 1.6.0 will be released?

I'm busy working on a more stable Docker solution than what I have now and having the latest and greatest version would be supreme!

MartinH0:
@KirkBushman is there any feedback on the issues @boban-dj found?
If you guys can discuss this Issues [HERE](https://community.invoiceplane.com/t/topic/8544) would be awesome.

After all issues are fixed this will be implemented into core.

KirkBushman:
I'll look at it tomorrow, follow you up on that

boban-dj:
Thanks Kirk!

MartinH0:
@KirkBushman pls also read his [POST](https://community.invoiceplane.com/t/topic/8544/26?u=martin_anonym) in the Forum as it shows 4 issues (2 more then here)

boban-dj:
@MartinH0, not sure where to post so please correct me if the ip forum is the way to go.

in `application/modules/invoices/models/Mdl_item_amounts.php`

`public function calculate($item_id)` is checking for `get_setting('default_include_item_tax') === "0"` but its

		var_dump(get_setting("default_include_item_tax"))
		string(0) ""

if you leave empty 
	
		`get_setting('default_include_item_tax') === ""`

it calculates the item tax rates with discounts(amount and percent) correctly.
But then again the subtotal (in totals columns below product rows) is not calculated with the correct discounted values. Which leads to issue I described earlier: if you use item tax and item discount, and add invoice discount which unit tax to alter? That is also my argument for keeping the 2 taxes separate. In bookeeping it makes sense and is done like this why not in code?

I would love to change this, but then again there must be an agreement about the principle I stated earlier.
You have only item tax OR invoice tax. So then this becomes a feature (and is needed in the real world)
Then we would have more clean code and usability for the users.

**added screenshots from this change**
removed_0

MartinH0:
@boban-dj yes pls post & discuss in IP Forum.
Copy your post over to there and lets discuss it here. This is more for technical things. Not discussions.

But anyway, thanks for your effort of testing and analysing it.

KirkBushman:
Answered on the forum

noplanman:
Kind of on topic, where does the active development take place?

https://development.invoiceplane.com/ seems to be down, so wondering where it has moved to.

clockwiseq:
> Kind of on topic, where does the active development take place?
> 
> https://development.invoiceplane.com/ seems to be down, so wondering where it has moved to.

Apologies, the certificate expired and we are working to get that resolved.

noplanman:
Hi all, just wondering if there is any progress on this 😇

Issue #734, opened 2021-06-12
CVE-2021-29022: the upload feature discloses the full path of the file upload directory


zeitschlag:
> In InvoicePlane 1.5.11, the upload feature discloses the full path of the file upload directory

- [cve.mitre.org](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-29022)
- [Exploit](https://notnnor.github.io/research/2021/03/17/full-path-discloure-in-invoiceplane.html)

zeitschlag:
The idea is just to remove the full path, because it seems that it's not even used.

PR #739, opened 2021-06-13
Don't show upload-path

zeitschlag:
This is the quick fix for #734

As the full path is never used, we just decided to remove it entirely and not show it at all.


clockwiseq (approved):
Looks fine to me.  Reviewing the original code, I'm also not sure why the full path was included.

PR #805, opened 2022-03-10
Fix Payment Method Select

tridnguyen:
When a payment method exists on the invoice, the select is disabled with the wrong option selected.  Set the select value and trigger a change so that select2 re-renders the selected value.

Pull Request Checklist

  * [X ] My code follows the code formatting guidelines.
  * [ ] I have an issue ID for this pull request. 
  * [ ] I selected the corresponding branch.
  * [ ] I have rebased my changes on top of the corresponding branch.
  
  * [x ] Bugfix
  * [ ] New Feature

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ adminer.php
 /nbproject
 /.php_cs.cache
 /doc
+.history
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ adminer.php
 /nbproject
 /.php_cs.cache
 /doc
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@ adminer.php
 /nbproject
 /.php_cs.cache
 /doc
+.history
```
UnderDogg, line 18: @tridnguyen thanks for the PR, man! It looks good.

Why did you add this to the .gitignore file?

Issue #879, opened 2022-11-04
Display clients list ordered by surname


giacy86:
As title.

giacy86:
https://github.com/InvoicePlane/InvoicePlane/pull/880

PR #880, opened 2022-11-04
[IP-879]: Order clients by surnames

giacy86:
Pull Request Checklist

  * [ ] Bugfix
  * [ ] New Feature


giacy86:
https://github.com/InvoicePlane/InvoicePlane/issues/879

fabm3n:
Hi @giacy86 ,

i don't think it will be a good idea to just change the default behaviour, because the surname is an optional field and therefore can be empty.
It would be better to make the client overview more flexible to order all columns like everyone likes.

PR #891, opened 2022-12-01
MySQL 8 compatibility

nielsdrost7:
Improves compatibility with MySQL version 8 by comparing to NULL dates instead of '0000-00-00'


nielsdrost7 (commented):
Approved

Issue #914, opened 2023-02-03
$item->product_sku in invoices but not in quotes?


naui95:
For the issue see: https://community.invoiceplane.com/t/topic/10126

See fix here: https://github.com/InvoicePlane/InvoicePlane/pull/717 (solution by @ctsde-markus)

To-do
- [ ] test against current release

nielsdrost7:
@ctsde-markus we'll make sure to merge your patch this time

naui95:
Will close the issue since @nielsdrost7 merged. Thanks @ctsde-markus for the fix and the heads up.

PR #717, opened 2021-01-15
Make product.* available in quotes

ctsde-markus:
Make $item->product values available in quotes (just as they are in invoices) as announced in the forum https://community.invoiceplane.com/t/topic/9188

Pull Request Checklist

  * [x] My code follows the code formatting guidelines.
  * [ ] I have an issue ID for this pull request. (Issues are deactivated?)
  * [x] I selected the corresponding branch.
  * [x] I have rebased my changes on top of the corresponding branch.
  
Issue Type (Please check one or more)

  * [ ] Bugfix
  * [x] Improvement of an existing Feature 
  * [ ] New Feature

PR #915, opened 2023-02-03
Make product.* available in quotes

nielsdrost7:
This patch never made it to production last time, so now we're making sure it will

Refs #914 supersedes #717

> Here is my patch to make values from the product table also available in the pdf template for quotes:


naui95 (approved):

nielsdrost7:
Thanks @ctsde-markus , it's merged in the Beta release, which means it will make it to 1.6.1 final.
We'll give you credit in the readme, once 1.6.1 final comes out :fist_oncoming:

Issue #959, opened 2023-07-09
After installing InvoicePlane, the ENCRYPTION_KEY isn't set


nielsdrost7:
## Expected Behavior
After installing InvoicePlane, the ENCRYPTION_KEY should be set

## Current Behavior
After installing InvoicePlane, the ENCRYPTION_KEY isn't set

## Possible Solution
Make sure that during installation, in the SetupController, we set the ENCRYPTION_KEY

## Steps to Reproduce
1. Clean install of InvoicePlane
2. Before installing, check the ipconfig.php file
3. Value of Encryption Key is: `ENCRYPTION_KEY=`
4. After installing, Value of Encryption Key is: `ENCRYPTION_KEY=`

## Possible Implementation
Make sure that during installation, in the SetupController, we set the ENCRYPTION_KEY

PR #960, opened 2023-07-09
[IP-959]: set encryption key after installing

nielsdrost7:
## Description
After installing InvoicePlane, the ENCRYPTION_KEY should be set

## Related Issue
#959 

## Steps to Reproduce
1. Clean install of InvoicePlane
2. Before installing, check the ipconfig.php file
3. Value of Encryption Key is: `ENCRYPTION_KEY=`
4. After installing, Value of Encryption Key is: `ENCRYPTION_KEY=`

  * [x] Bugfix


clockwiseq (approved):
Looks good to me.

Issue #978, opened 2023-09-10
Can not overwrite SESS_MATCH_IP


musapinar:
## Expected Behavior
I expect the app not to log out a user when his ip changes

## Current Behavior
User is redirected to the authentication form if his ip becomes different from the one during the session creation.

## Screenshots
-

## Possible Solution
-

## Steps to Reproduce
1. Grab a smartphone
2. Connect to your home wifi network
3. Log in to your InvoicePlane app and browse the app
4. Turn off wifi and connect to your cellular network instead (5g or whatever)
5. Browse your app, notice you have been logged out and redirected to the authentication form

## Context (Environment)
User browsing the app behind a mobile network, typically via a smartphone, away from home, often have a non-static ip. The ip may change every few minutes depending on your country/isp
User may have ip 1.2.3.4 the moment he opened the invoice creation form page, but ip 4.3.2.1 the moment he submitted the completed form. This will lead to a failure of the ajax request when saving the invoice. User is frustrated because the page has to be refreshed (because he has to re-authenticate), so data is completely lost.

## Detailed Description
-

## Possible Implementation
The quickfix I went for was to put ```SESS_MATCH_IP=false``` within ipconfig.php
But this did not work until I had to replace ```$config['sess_match_ip'] = env('SESS_MATCH_IP', true);``` to ```$config['sess_match_ip'] = env_bool('SESS_MATCH_IP', true);``` here https://github.com/InvoicePlane/InvoicePlane/blob/development/application/config/config.php#L430

## Failure Logs
-

nielsdrost7:
Ok, I'll put your fix in the upcoming release

PR #982, opened 2023-09-17
#978: When on a mobile the SESS_MATCH_IP environment variable is important. Changed a setting in /application/config.php to env_bool instead of _env

nielsdrost7:


nielsdrost7:
@musapinar I've made the change you suggested in issue #978. Feel free to review it.

PR #1007, opened 2023-12-14
[IP-1003]: Add extra field title

KevinJ30:
- Add extra field title
- Simplification conditional
#1059


nielsdrost7:
Merged, will be released in version 1.6.2

Issue #1033, opened 2024-02-25
Creation dates of invoices should never be changed


mheiduk:
Based on a conversation in another issue with @Verony-makesIT: https://github.com/InvoicePlane/InvoicePlane/issues/1025#issuecomment-1919883551

Creation dates of invoices should never be changed and this feature first mentioned in https://github.com/InvoicePlane/InvoicePlane/issues/895 should just change the due date. This is highly confusing when the creation date invoices are changing, especially when they are created eg. in 2023 and a reminder will be sent out in 2024.

I know that you can disable it, but 1) this is the default behaviour if you setup invoiceplane out of the box and 2) it is not legal to do that. The correct way to change an invoice creation date is to cancel the invoice and create a new one.

I can provide a suitable pull request if needed.

nielsdrost7:
Don't worry about the PR for now.
I don't think I've merged that PR, at least, I don't think so 🤔.
You are absolutely right about that invoice date.
If we won't change anything there, we won't have a problem, correct?

novafirst:
Apologies for the interruption. Here are my thoughts:

Invoice Information:
- Invoice Created Date/Time: 09-20-2023 17:14:08
- Invoice Updated Date/Time: 09-21-2023 11:11:34
- Invoice Date: 09-23-2023
- Invoice Due Date: 10-04-2023

Please note that the first two fields, "Invoice Created Date/Time" and "Invoice Updated Date/Time," should be for internal reference only and should not be printed or displayed on the invoice itself. These timestamps should automatically be generated and retained solely for internal auditing purposes.

Regarding the last two fields, "Invoice Date" and "Invoice Due Date," these are the crucial details to be presented on the invoice. The "Invoice Date" should be selectable by the user, ensuring it falls between the current date and the "Invoice Due Date." It's important to consider that payment terms may vary; some invoices may require payment upon receipt, while others may follow specific terms like NET 10, 15, 30, etc.

mheiduk:
@Verony-makesIT I understand your point that when you initially send the invoice to the client, the creation date can be changed to the date when you've sent the invoice. But adjusting the creation date when you send the client a reminder email is confusing and should not happen. You can circumvent this behaviour if the read-only setting is set to "Sent", but when it's set to "Paid" the reminder email will update the creation date nonetheless.

To prevent further confusion when to update the creation date and when not, the creation should not be updated automatically at all in my opinion. If you still want to change the creation date of the invoice, you can do it manually anyway.

PR #1036, opened 2024-03-03
[IP-1033]: Bugfix: creation dates of invoices should never be changed

mheiduk:
## Description
only update invoice_date_due when sending email

invoice_date_created should not be updated and always stay the same

See (https://github.com/InvoicePlane/InvoicePlane/issues/1033) for more details

## Pull Request Checklist

  * [x] Bugfix
  * [ ] New Feature


nielsdrost7:
Great PR.
How do I test proper working of this PR?
What do I fill where in order to see it in action?
What should happen?
What should not happen?

Verony-makesIT:
@mheiduk, 
Item #1036 can never be called a "bug fix" but rather, in my opinion, is a different interpretation or perspective of how **the invoice creation date _versus_ the invoice date** should be treated.
The result of your modification is that on your invoice, the invoice (creation) date may be (weeks or days) before the shipment date (official invoice date) and the due date may be more than 30 days later after that invoice date. 
This is very confusing for the customer. 
_For example:_
Period to due date = 30d
creation date: 27 dec 2023
work performed in January 2024
invoice updates: 10/01/2024 + 17/01/2024 + 28/01/2024
shipment date: 01/02/2024
_Result:_
Invoice date: **27/12/2023**
Due date: **01/03/2024**
Fiscal year: 4th quarter 2023 or 1st quarter 2024 (?)

I think tax authorities, your accountant and your client will be asking questions anyway....
Please also read my comments in item #1033.
**In any case, I cannot accept this.**

nielsdrost7:
@mheiduk again, thank you for this PR. I'll try to go over it this weekend.
If you could just answer those questions I asked, that would be awesome.

mheiduk:
@nielsdrost7 
> How do I test proper working of this PR? What do I fill where in order to see it in action? What should happen? What should not happen?
You have to send an overdue invoice to a client. If the setting in Settings -> Invoices -> Other settings is set to "No", then only the due date should update to the current date. If the setting is set to "Yes", then no date should be updated. The creation date should not be updated at all.

nielsdrost7:
Thanks mam! I'll take a look this weekend

application/language/english/ip_lang.php
```diff
@@ -332,7 +332,7 @@
     'no_overdue_invoices'                          => 'No overdue Invoices',
     'no_quotes_requiring_approval'                 => 'There are no quotes requiring approval.',
     'no_updates_available'                         => 'No updates available.',
-		'no_update_invoice_due_date_mail'							 => 'Disable the change of invoice date and due date before emailing',
+		'no_update_invoice_due_date_mail'							 => 'Disable the change of invoice due date before emailing',
```
nielsdrost7, line 335: Just the indenting, nothing special.
Also please align the arrow with the other arrows

nielsdrost7 (commented):
Also:
`public function mark_sent($invoice_id)` in `Mdl_invoices.php`
Shouldn't you check for that due date as well there?

Because right now it's only checked when the invoice is emailed, but when someone manually sets the invoice to `sent` that due_date isn't checked?

What do you think?

mheiduk:
> Also: `public function mark_sent($invoice_id)` in `Mdl_invoices.php` Shouldn't you check for that due date as well there?
> 
> Because right now it's only checked when the invoice is emailed, but when someone manually sets the invoice to `sent` that due_date isn't checked?
> 
> What do you think?

Thanks for the input! Tried to simplify it as much as possible and cleaned up some things in the process. Please check if everything is working as expected for you.

application/language/english/ip_lang.php
mheiduk: Should be fixed

nielsdrost7:
Great, great Job!
I merged it into our `1033` branch, as I now noticed and then merged the `1033` branch into `develop`.
I'll make sure to credit you, once we make the new release.

It was fun doing this PR with you.

Issue #757, opened 2021-09-16
Paypal API interaction with Client ID and Secret


ririko5834:
Please create PayPal API integration to allow us to use Client ID and Secret for Paypal interaction. As on normal personal PayPal accounts, you can't get PayPal express credentials. 
Here is PayPal API documentation https://developer.paypal.com/docs/api/reference/api-requests/

NoodlesDev987:
I have same problem that I can't get paypal express credentials

nielsdrost7:
See PR #816 , I'm closing this issue

anondev-sudo:
This wasn't fixed, just stripe integration, not paypal.

Also when there will be release?

nielsdrost7:
@anondev-sudo the release depends on a couple of things.
In the meantime you can check out the `development` branch and maybe merge in PR #831. It is prepared for PHP 8.1 compatibility.

anondev-sudo:
Can you reopen this? The PR was for stripe outdated fix, but not foe PayPal to use client id and secret new api. The express api is outdated and doesn't work on personal PayPal accounts while you can obtain client id and secret also on personal accounts to accept payments

nielsdrost7:
It's reopened. Hope you can make a pull-request for the fix.

naui95:
@anondev-sudo @ririko5834 are you available to make some testing? We set up a new payment interface that supports Paypal (and Stripe). We would be grateful if you guys could do some testing.

Issue #1032, opened 2024-02-23
[Feature Request] Mollie - Request Online Payment vendor


siniztr:
This is a request to add the Online Payment vendor Mollie. This is a very popular European payment vendor. The documentation for doing so can be found here: https://docs.mollie.com/
By adding this payment vendor, it opens users up to accepting many types of payments, such as bank transfers, paypal, apple and and google pay, as well as many EU countries instant bank transfers (Dutch: iDeal, Belgium:BanContact, etc.)

nielsdrost7:
Todo:
https://github.com/thephpleague/omnipay-mollie  :
`composer require league/omnipay omnipay/mollie`
Then use `basic purchase example` from the readme

nielsdrost7:
@siniztr would you like to help test this PR

rcarambot:
Can I put in a request for btcpay (https://btcpayserver.org/) as one of the choices for payment gateway?

ncafassassin:
Can I put in a request for btcpay (https://btcpayserver.org/) as one of the choices for payment gateway?

nielsdrost7:
@ncafassassin I've added the direct link to the documentation https://docs.btcpayserver.org/API/Greenfield/v1/ and since they have an API we should be able to connect with it, so yes, your request was added

naui95:
We are removing this feature from v1.6.2. Will be provided in further updates.

nielsdrost7:
@coderabbitai make plan

Issue #1006, opened 2023-12-04
Stripe payments erroring because of invalid content-type


jdanna:
## Expected Behavior
After adding stripe as a provider, attempting to pay an invoice results in an content-type error from stripe.
## Screenshots
![IMG_3670](https://github.com/InvoicePlane/InvoicePlane/assets/25439364/fa0cd9a9-cd39-451a-89fa-2b47ebf476fc)
![IMG_3671](https://github.com/InvoicePlane/InvoicePlane/assets/25439364/ae0c5f67-fe9a-4c7c-b92f-8e5b3b4119b5)
![IMG_3672](https://github.com/InvoicePlane/InvoicePlane/assets/25439364/d42cdc77-2172-4509-b2b8-3a80c39c06fa)

Stripe shows errors happened, but doesn’t provide any more detailed logs

## Possible Solution

Looks like its related to how omnipay is making the request:

https://stackoverflow.com/questions/66241686/laravel-omnipay-stripe-unable-to-send-request-on-linux/66290005#66290005

I tried modifying the payment handler php to add the headers, but it didn’t work

## Steps to Reproduce
-Enable online payments
-Enter stripe keys and enable stripe (i tested with both test mode and live mode keys in stripe - both had the same issue)
-Send an invoice
-Try to pay the invoice with a credit card

nielsdrost7:
@jdanna great news! A pull-request is on its way to resolve this issue.
Once it's done we'll let you know

naui95:
@jdanna are you available to test the new feature? We set up a new payment interface that supports Stripe (and PayPal). We would be grateful if you could do some testing.

idressos:
Experiencing the exact same issue with InvoicePlane v1.6.1 - PHP8.1-FPM - Debian 12.

nielsdrost7:
@idressos see the solution in this branch:
https://github.com/InvoicePlane/InvoicePlane/pull/1046
Are you available to test the branch?

idressos:
> @idressos see the solution in this branch: #1046 Are you available to test the branch?

Yes, I'm available. Please provide any further instructions if necessary except for cloning and deploying.

nielsdrost7:
I _think_ these would be the steps:

-Enable online payments
-Enter stripe keys and enable stripe
-Send an invoice
-Try to pay the invoice with a credit card

idressos:
> I _think_ these would be the steps:
> 
[...]

PR #1046, opened 2024-04-28
[IP-1006]: feature: payments v1.6.2

naui95:
## Description
Implementing an overhauled payment system that supports out of the box PayPal, Stripe, Stripe Checkout. Mollie (#1032) will also be supported.

## Related Issue
This PR closes the issues #757 #1006 

## Motivation and Context
The implementation of omnipay in InvoicePlane has become a bit tricky and a better and more stable solution wants to be provided.


nielsdrost7:
Thank you so much for this PR! Tiny review incoming

nielsdrost7 (commented):
Open `/application/modules/settings/views/partial_settings_invoices.php` (you didn't change this file, but I would love a tiny change):

Find this piece of code:
```
<select name="settings[gateway_<?php echo $d; ?>_currency]"
```
It contains the selectbox for the _gateway_currencies_ for the payment provider.
On our `General` tab in de settings we have this setting (I'll add it with a `get_setting`, since you're going to need that one:
`get_setting('currency_symbol');`

So my request:
Can you in that selectbox for _gateway_currenctes_ match the selected setting for `get_setting('currency_symbol');`?

That way you'll get a selected USD as `gateway_currency` instead of the goold ol' `AED` as currency symbol for a `gateway_currency`.

Issue #1059, opened 2024-06-22
[Feature Request] Text field for Mr. and Ms. in customer section


lxwulf:
It's common to use `Mr.` and `Ms` in invoices.

**Private customer example:**

```text
Mr.
Johannes Gutenberg
Bookstreet 42
55116 Mainz (a Rhein)
```

**Business customer example:**

```text
Printing Press Gutenberg Inc.
Mr. Johannes Gutenberg
Bookstreet 45
55116 Mainz (a Rhein)
```

**At the moment:**

```text
Johannes Gutenberg
Bookstreet 42
55116 Mainz (a Rhein)
```
It would be one additional field above the normal `name` field for the `private customer` example. 
On company side, it's common to use the addressing/salutation directly before the personal name.

nielsdrost7:
refs #1007

nielsdrost7:
@lxwulf than you for your issue. It's a duplicate of #1003 , but I'll keep your feature request as well.
See the PR I attached. It's in `draft`, but if you want you can test if out.
It has a new field in that pull-request do your database will change just a tiny bit.
Just so you know.

lxwulf:
Thank you for the information and sorry I missed the mentioned issue! 
I have extra set up a dev server, so I can test exactly such things. :blush:

> Sorry for asking, but how can I test this commit if its in draft? I did not saw any branch with this changes. 
> Or can I also just checkout this commit?
> :point_right: :point_left:

nielsdrost7:
@lxwulf see Pull-Request https://github.com/InvoicePlane/InvoicePlane/pull/1101
it's regarding adding a title for a client.
- `git clone invoiceplane/invoiceplane`
- `git checkout feature/1059-branch`
- just install in the ususal way (`composer install, yarn install, yarn grunt`)

nielsdrost7:
Merged, will be released in version 1.6.2

Issue #1003, opened 2023-11-29
[Feature Request] Extra field "title" for personal client information


onny:
It would be nice to have an extra field in the personal information area of a client for specifying a title.
This should get rendered in the invoice pdf in the billing to area:

Mr.
Christian Bauer
Example Street 1
Example City
etc.

The extra field could be a drop down for common titles like "Mr" or "Mrs" but also should provide a free form text field in case one would need to specify a complex title (including academic title etc).

nielsdrost7:
So a select box with a list with contents:
[array of settings like Mr., Mrs, etc] AND a list-item: Custom (or 'type it yourself).

Save the chosen title to "create field in `clients` table

nielsdrost7:
Yes we can, but it would still be a select box with that list of possibilities.
Plus I know someone who wants to manipulate it through the ipconfig file.

@KevinJ30 would you like to make a branch with the suggested solution? For example with your suggestion of that enum you were talking about

nielsdrost7:
@onny I've improved the Pull-Request that @KevinJ30 made. Could you take a look?
https://github.com/InvoicePlane/InvoicePlane/pull/1101

nielsdrost7:
@onny Merged, will be released in version 1.6.2
can you check please in the Beta version (coming up soon) if this is what you meant?

onny:
Thank you @nielsdrost7, going to test it today. Just reworked the InvoicePlane package in NixOS so it will be easier to test development branches https://github.com/NixOS/nixpkgs/pull/353783 :)

nielsdrost7:
> Just reworked the InvoicePlane package in NixOS so it will be easier to test development branches

That's awesome, man.
By the way: I just thought of something.
Would you like the values in the ClientTitle to be configurable?
Maybe a setting in ipconfig.php?
Comma separated, lowercase?

If my colleagues won

onny:
Just tried the new feature, seems to have some issues: After clicking on "Add client" I get: 

``` 

An uncaught Exception was encountered

Type: TypeError

Message: Clients::{closure}(): Argument #1 ($clientTitleEnum) must be of type ClientTitleEnum, stdClass given

Filename: /nix/store/ngx36v3bsk8cxwdngcxbcc0rb7v34jbl-invoiceplane-localhost-1.6.1/application/modules/clients/controllers/Clients.php

Line Number: 272

Backtrace:

[...]

PR #1101, opened 2024-10-30
[IP-1003]: Add extra field title

nielsdrost7:
- Add extra field title
- Simplification conditional
#1059 
#1003

### DIFF
--- a/application/modules/clients/Enums/ClientTitleEnum.php
+++ b/application/modules/clients/Enums/ClientTitleEnum.php
@@ -1,4 +1,4 @@
-<?php declare(strict_types=1);
+<?php
 
 enum ClientTitleEnum: string
 {
--- a/application/modules/clients/Enums/ClientTitleEnum.php
+++ b/application/modules/clients/Enums/ClientTitleEnum.php
@@ -1,0 +1,10 @@
+<?php declare(strict_types=1);
+
+enum ClientTitleEnum: string
+{
+    case MISTER = 'mr';
+    case MISSUS = 'mrs';
+    case DOCTOR = 'doctor';
+    case PROFESSOR = 'professor';
+    case CUSTOM = 'custom';
+}
--- a/application/modules/clients/Enums/ClientTitleEnum.php
+++ b/application/modules/clients/Enums/ClientTitleEnum.php
@@ -1,10 +1,57 @@
 <?php
 
-enum ClientTitleEnum: string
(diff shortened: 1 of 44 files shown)


application/modules/clients/controllers/Clients.php
```diff
@@ -253,4 +267,12 @@ public function delete($client_id)
         $this->mdl_clients->delete($client_id);
         redirect('clients');
     }
+
+    private function get_client_title_choices(): array
+    {
+        return array_map(
```
nielsdrost7, line 273: @naui95 
so if we alllow fn functions, we cannot support PHP 7.4 anymore, correct?

@KevinJ30 I really like this function man, kudos!

application/modules/clients/models/Mdl_clients.php
```diff
@@ -35,93 +35,97 @@ public function default_order_by()
 
     public function validation_rules()
     {
-        return array(
-            'client_name' => array(
+        return [
+            'client_title' => [
+                'field' => 'client_title',
+                'label' => trans('client_title'),
```
nielsdrost7, line 41: @KevinJ30 I removed the "required" part from the validation.
When you enter a new client the "title" isn't required

application/modules/clients/controllers/Clients.php
naui95: @nielsdrost7 `fn` functions have been introduced in PHP 7.4 so we should be ok with them. Starting from IP1.6 we also state that minimum PHP version is 8.0 so this should not be an issue.

application/modules/clients/Enums/ClientTitleEnum.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+enum ClientTitleEnum: string
```
naui95, line 3: @nielsdrost7 the solution provided is cool, but an `enum` will break backward compatibility. Enums were introduced in PHP 8.1 (https://www.php.net/manual/en/language.types.enumerations.php). Can we find an alternative solution?

naui95 (changes requested):

nielsdrost7:
@KevinJ30 i moved the block for client_title to the bottom of the page (near Gender)
For the rest: awesome PR man!

naui95 (approved):

nielsdrost7:
@KevinJ30 Merged, will be released in version 1.6.2

PR #1105, opened 2024-11-05
Custom template with zero values emptied

nielsdrost7:
Custom template with zero values emptied


nielsdrost7:
We're not adding new templates to the base InvoicePlane. We'll add it to the Templates section on the forums.

I've asked however to look at the empty values, since that was the original question

naui95:
> Where to find the community provided templates?

Community provided templates and themes can be found at www.invoiceplane.com/themes you are welcome to share your solutions!

Issue #1130, opened 2024-12-05
Error on input validation (1.6.2 beta)


xam-ps:
1. Go to any input form (e.g. add new tax).
2. Click on save without putting in any values

## Expected Behavior
An input validation should be shown to the user.

## Current Behavior
The application tries to save the values and is running into an error after some time.

## Screenshots
![grafik](https://github.com/user-attachments/assets/73f06f54-d9ff-4ff1-8d95-6aed2bfb2c03)

## Steps to Reproduce
1. Go to any input form (e.g. add new tax).
2. Click on save without putting in any values

## Detailed Description
This happens on any form
- custom_fields form
- email_template form
- invoice_groups/form
- etc.

nielsdrost7:
@sudwebdesign could you take a look at this one as well, please?

sudwebdesign:
Hi, @nielsdrost7 
The fix #1155 by @AeroBytesNL should be fine :+1:.
Note:
Only remove `console.log('ajax-loader');` at line [249](https://github.com/InvoicePlane/InvoicePlane/pull/1155/files#diff-2cb46051b408c1dda73ebd6f22c3ba84076bebdc790ba80630b75d8ca5c790c2R249) to be perfect.

AeroBytesNL:
@sudwebdesign @nielsdrost7 i removed the `console.log()` :)

nielsdrost7:
Solved by #1155 , but I'll keep this one open till after the release

nielsdrost7:
Solved by [releasing version 1.6.2](https://github.com/InvoicePlane/InvoicePlane/releases/tag/v1.6.2)

PR #1179, opened 2025-01-03
Fix #fullpage-loader (Spinner) never showed after save

sudwebdesign:
## Description

`$(document).on('click', '.ajax-loader', function () {` duplicated inside same function.

## Pull Request Checklist

  * [ ] I have an issue ID for this pull request.

  * [x] Bugfix
  * [ ] New Feature

### DIFF
--- a/assets/core/js/scripts.js
+++ b/assets/core/js/scripts.js
@@ -225,8 +225,8 @@ $(document).ready(function () {
         update_email_template_preview();
     });
 
-    // Fullpage loader
-    $(document).on('click', '.ajax-loader', function (event) {
+    // Fullpage loader (spinner)
+    $(document).on('click', '.ajax-loader', function () {
         var requiredFieldsFilledIn = true;
 
         $('input[required], textarea[required], select[required]').each(function () {
@@ -235,9 +235,9 @@ $(document).ready(function () {
             }
         });
 
-        // Checks if required fields are filled it. If not, don't display spinner.
+        // Checks if required fields are filled it.
         if (!requiredFieldsFilledIn) {
-            return;
+            return; // If not, don't display spinner.
         }
 
         // Show loader
--- a/assets/core/js/scripts.js
+++ b/assets/core/js/scripts.js
@@ -240,13 +240,13 @@ $(document).ready(function () {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


nielsdrost7:
@sudwebdesign I think we had that because of this problem: https://github.com/InvoicePlane/InvoicePlane/issues/1130

- Apply your fix
- Leave forms like the product form or payment form completely empty and save

The full page loader will show, but it will **time out** and it will not show the validation errors.

If you can resolve that problem, I'll be very grateful.

(The bug was introduced between 1.6.1 and 1.6.2)

sudwebdesign:
> The full page loader will show, but it will time out and it will not show the validation errors.

Sorry but no, this is solved by [`if (!requiredFieldsFilledIn)`](https://github.com/InvoicePlane/InvoicePlane/pull/1179/files#diff-2cb46051b408c1dda73ebd6f22c3ba84076bebdc790ba80630b75d8ca5c790c2L239)  [see](https://github.com/InvoicePlane/InvoicePlane/pull/1155/files#diff-2cb46051b408c1dda73ebd6f22c3ba84076bebdc790ba80630b75d8ca5c790c2R239) in #1155

But the 2th event (in event) never called and never show `#fullpage-loader`, this why exist this Fix.

Proof by img: 

required-fields-ok


(After add name, i push the `Enter` and go to empty `Price` field, After set and push `Enter` or Click `Save` button show `#fullpage-loader` normally.

sudwebdesign:
@nielsdrost7 (maybe after apply this patch) need to reload browser cache with `Ctrl + F5`

sudwebdesign:
Hi @nielsdrost7 

Before the fix, need to click twice on `Save` button to show the spinner (double send request on server is possible)

> I think we had that because of this problem: https://github.com/InvoicePlane/InvoicePlane/issues/1130

Maybe it's an oversight when editing the code, it doesn't make sense to have a click event within a click event. I missed the double event in PR #1155, mea culpa.

The problem with _required fields_ #1130 already been corrected by #1155

sudwebdesign:
It's work with new client, product, ... (when element have `ajax-loader` in attribute class)

required-fields-client-ok

nielsdrost7:
@sudwebdesign merci!
I'll take a look when I have some time

PR #1236, opened 2025-02-22
Add page numbers to PDF footer

### LINKED ISSUE #1234: Pagination in invoice php template
### Title

_No response_

### Feature Description

Hey, would be nice to have pagination for the invoice template and reference page numbers somehow.

### Use Case

_No response_

### Additional Context (Optional)

_No response_

### Priority Level

None

pekiZG:
## Description

Added page numbering to the bottom of the PDF.

## Related Issue

https://github.com/InvoicePlane/InvoicePlane/issues/1234

## Motivation and Context

Page numbers were requested by @onny in #1234

## Screenshots (if appropriate):
image
image

## Pull Request Checklist

  * [ ] Bugfix
  * [x] New Feature

## Summary by CodeRabbit

* **New Features**
  * Added a toggle to include page numbers in generated PDFs.

* **Chores**
  * Added/updated many CI workflows for testing, linting, dependency updates, releases, and translation sync; added workflow documentation.
  * Updated .gitignore.

* **Refactor**
  * Widespread code formatting and style cleanups.

* **Improvements**
  * Strengthened filename, referer/same-origin handling and related safety checks; enhanced bot-detection heuristics.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


nielsdrost7 (changes requested):
@copilot
- Around line 50-58: The current workflow step "Run Composer audit" uses a file
size check on audit-report.json which is a false positive because composer audit
emits {"advisories": {}} even when no issues exist; change the check to parse
the JSON and inspect the "advisories" object instead (for example, use jq or a
short python snippet to test if .advisories is empty) and set the GITHUB_OUTPUT
variable "vulnerabilities_found" to true only when .advisories contains entries;
update the step with the id "audit" to replace the -s audit-report.json
conditional with a JSON-aware check against .advisories.

In @.github/workflows/phpstan.yml:
- Around line 65-66: The "Comment on PR (if available)" step has an unreachable
condition (github.event_name == 'pull_request') because this workflow only
triggers on workflow_dispatch; update the workflow trigger or the step
condition: either add pull_request to the top-level on: triggers in this
phpstan.yml so the condition can be true, or change/remove the step's if: guard
and instead use a trigger/condition that matches how you intend to run it (e.g.,
use workflow_run or check github.event_name == 'workflow_run' / pass a PR
context from a caller) so the step can actually execute; locate the step by its
name "Comment on PR (if available)" to apply the change.

In @.github/workflows/phpunit.yml:
- Around line 37-49: The workflow contains Laravel-specific steps (php artisan
key:generate, php artisan test) and references .env.testing.example which are
not applicable to this CodeIgniter 3.3 project; remove or replace those lines
and instead run PHPUnit directly (e.g., vendor/bin/phpunit or phpunit) with the
repository's phpunit.xml configuration, ensure tests use the project's
configuration file (ipconfig.php) by copying or linking it into the test
environment before running tests, and delete any references to artisan commands
(php artisan key:generate, php artisan migrate, php artisan db:seed) so the CI
either runs CodeIgniter-compatible PHPUnit steps or the workflow is removed if
tests are not intended to run.

In @.github/workflows/pint.yml:
- Around line 93-101: The "Commit and Push Changes" workflow step currently
always runs git push and will fail for forked PRs because GITHUB_TOKEN is
read-only; update the step's if condition (the step named "Commit and Push
Changes" that already checks steps.verify_diff.outputs.has_changes) to also
require that the run is not a forked pull request by validating same-repo or
non-pull_request contexts (for example, require
github.event.pull_request.head.repo.full_name == github.repository or require
github.event_name != 'pull_request' / github.ref startsWith 'refs/heads/' for
manual/branch runs) so the push only runs when the action has push permission.

In @.github/workflows/quickstart.yml:
- Around line 64-65: The workflow currently runs the "Run database migrations"
step which invokes `php artisan migrate --force` unconditionally and will fail
without a DB; either remove or comment out that step (the step named "Run
database migrations" invoking `php artisan migrate --force`) if smoke tests
don't need DB, or alternatively ensure tests use an in-memory SQLite by updating
`.env.testing.example` to set DB_CONNECTION=sqlite and DB_DATABASE=:memory:, or
re-enable the MySQL service in the workflow so migrations can run; pick one
approach and apply it consistently so the `php artisan migrate --force` step no
longer fails.

- Around line 425-439: The workflow step "Create GitHub Draft Pre-Release"
currently uses softprops/action-gh-release@v1 which is deprecated; update the
action reference to the v2 release (e.g., softprops/action-gh-release@v2 or
`@v2.5.0`) in the step that defines tag_name/name/body_path/files/draft/prerelease
so the newer action is used; after changing the version, verify input names
(tag_name, name, body_path, files, draft, prerelease) still match v2 docs and
adjust any differences in parameter names or behavior for the step if needed.

In @.github/workflows/test-php.yml:
- Around line 3-4: The workflow trigger was changed to only workflow_dispatch
which disables automatic PR testing; restore automatic runs by adding the
pull_request trigger back alongside workflow_dispatch (e.g., include a top-level
triggers block that contains both pull_request and workflow_dispatch) so tests
run on PR open/update while still allowing manual dispatch; update the
workflow's on: stanza to include both triggers.

- Around line 107-131: The security-only upgrade step currently uses the command
referenced as yarn upgrade "$pkg" --latest which can bump across major versions;
change it to use yarn upgrade "$pkg" (remove --latest) so upgrades respect
existing semver ranges in package.json and avoid breaking major-version jumps
for the packages gathered in VULNERABLE_PACKAGES; ensure the loop that echoes
and calls yarn upgrade uses the updated command.

- Line 87: The code uses mb_rtrim (e.g., when building $realBaseWithSep in
file_security_helper.php) which exists only in PHP 8.4+, causing fatal errors on
the declared PHP 8.1+ target; replace mb_rtrim with the multibyte-safe rtrim (or
call rtrim with a UTF-8-aware normalization step) so the code runs on PHP
8.1–8.3, and apply the same replacement to other occurrences (index.php lines
around 209/248/275, phpmailer_helper.php usage, pdf_helper.php,
dropzone_helper.php) or alternatively update the project minimum PHP requirement
and Setup.php PHP version check to 8.4.0 to match mb_rtrim usage.

In `@application/helpers/mailer_helper.php`:
- Around line 248-260: Replace the brittle str_starts_with() origin check with a
same-origin comparison that parses and compares URL components: call parse_url()
on both base_url() and $_SERVER['HTTP_REFERER'], ensure scheme, host and port
(normalize default ports) match before assigning $redirect = $referer, and treat
root-relative referers (starting with "/") as safe; otherwise set $redirect =
base_url(); apply the same fix where str_starts_with is used (mailer_helper.php
logic around $_SERVER['HTTP_REFERER'] and the Sessions controller check) and
keep the final redirect($redirect) behavior unchanged.

- Around line 20-31: The tryFrom method mis-handles array_search results and
returns the index instead of the matched string: change the truthy check to a
strict !== false (i.e. if ($searchResult !== false)) and set the returned
object's value to the actual string from getValues() (e.g. $returnObject->value
= $values[$searchResult] or simply $value) so lookups like 'mr' (index 0)
succeed and callers get the string value; keep the StdClass return pattern and
leave other behavior untouched.

PR #1238, opened 2025-03-05
[WIP] Add initial implementation of PDF417Barcode

### LINKED ISSUE #1017: [Feature request] Add option to add PDF417 barcode to invoice
## Expected Behavior
I would like to have an option to add a PDF417 barcode to the invoice which is used in Croatian invoices.

PDF417 spec can be found here (in Croatian language): https://www.hub.hr/sites/default/files/inline-files/2DBK_EUR_Uputa_1.pdf

## Screenshots
Example of an invoice with PDF417 barcode:
image

## Steps to Reproduce

1.
2.
3.
4.

pekiZG:
## Related Issue

https://github.com/InvoicePlane/InvoicePlane/issues/1017

## Motivation and Context

This implementation is being used in Croatia.

## Screenshots (if appropriate):
There are screenshots in liked issue.
image

## Pull Request Checklist

  * [ ] Bugfix
  * [x] New Feature

## Summary by CodeRabbit

* **New Features**
  * Added PDF417 barcode support for invoices in Croatia, providing an alternative barcode format compliant with local payment standards.

* **Bug Fixes**
  * Enhanced input sanitization to prevent injection vulnerabilities; improved consistency in logging for edge cases.

* **Documentation**
  * Added security and code quality analysis documentation.


xam-ps:
@pekiZG 
Looks good 👍 Just one question: Are you sure, that all people from Croatia prefer the 417 barcode over the QR-Code? Your PR will make it impossible for any person with a customer in Croatia to show a QR-Code on the invoice, won't it?

Also: This will only be displayed, when the QR-Code setting is enabled, but the 417 barcode is not really a QR-Code. Could maybe be confusing for some users.

But that's just my thoughts on this 😅

pekiZG:
Here are some options screenshots :)
image

And I just realised that smart thing should be to combine it into one.

sudwebdesign:
Intrigued by the `ip_lang` file (three leading spaces `   ` before `=>` numerous lines)
& composer.lock` restored

It simplifies the merger (and review readability)

See: https://github.com/InvoicePlane/InvoicePlane/pull/1238/commits/b155d49cce5fe9ce93b22c4e6caf21f8bc747061 and https://github.com/InvoicePlane/InvoicePlane/pull/1238/commits/3c28bf2bddcb9707a07c7e8cf8684892cb2bfe84

nielsdrost7:
> Intrigued by the ip_lang file (three leading spaces     before => numerous line

I'm intrigued by something else.
I see thr language code 'HR' hard-coded
So the rest of the world can't have bar-codes then?

Those spaces before the `=>` are there to align them under each other (I'm on mobile, so I cannot doublecheck)

It's a formatting rule in "**Pint**", a tool we use for our source code:

```
"operators": {
                "=": "align_single_space_minimal",
                "=>": "align_single_space_minimal"
            }
```

`vendor/bin/pint` will optimize the source code, however I usually apply it within my IDE when I'm working on a file.

nielsdrost7 (changes requested):
@copilot
- Around line 77-84: The current "Check for changes" step (id: check-changes)
only diffs composer.lock, so updates to composer.json are missed; change the git
diff invocation to include both files (composer.lock and composer.json) when
checking for quiet differences (e.g., git diff --quiet composer.lock
composer.json) and keep the same echo logic that writes changes_detected to
$GITHUB_OUTPUT so the step detects changes in either file.
- Around line 50-58: The workflow currently checks audit-report.json size to set
vulnerabilities_found, which is wrong because composer audit JSON is non-empty
even with no advisories; update the "Run Composer audit" step (id: audit) to
parse audit-report.json and inspect the "advisories" array instead of using -s:
run composer audit --format=json > audit-report.json, then use a JSON parser (jq
or node/python) to check whether .advisories has any entries and set the
GITHUB_OUTPUT variable vulnerabilities_found=true only if the advisories array
length > 0 (otherwise set vulnerabilities_found=false).

In @.github/workflows/docker.yml:
- Around line 23-27: The workflow uses the removed docker-compose v1 binary;
replace usages of the `docker-compose` commands in the "Build and start
containers (if docker-compose.yml is ready)" and "Check running services" steps
with the Docker CLI v2 form `docker compose` (e.g., change `docker-compose up -d
--build` to `docker compose up -d --build` and `docker-compose ps` to `docker
compose ps`) so the Actions runner uses the supported Docker Compose v2.

In @.github/workflows/phpstan.yml:
- Around line 65-66: The "Comment on PR (if available)" step is unreachable
because the workflow only has workflow_dispatch and the step condition uses
github.event_name == 'pull_request'; either add a pull_request trigger to the
workflow's on: block or remove/modify this step and its if condition. Locate the
step named "Comment on PR (if available)" in .github/workflows/phpstan.yml and
either add pull_request to the workflow triggers (so the step can run) or remove
the PR-only step/condition if you intend to keep workflow_dispatch-only runs.

In @.github/workflows/pint.yml:
- Around line 93-101: The "Commit and Push Changes" workflow step currently runs
git push on any PR with changes and will fail for forked PRs; update the step
condition (the step named "Commit and Push Changes" that currently checks
steps.verify_diff.outputs.has_changes) to also ensure the PR head repo matches
the target repo (e.g., add a guard using
github.event.pull_request.head.repo.full_name == github.repository or compare
github.event.pull_request.head.repo.owner == github.repository_owner) so the
push only runs for same-repo PRs (and still only when has_changes is 'true').
- Around line 40-63: The current use of vendor/bin/pint --dirty in the pint_run
step doesn't work in CI because actions/checkout produces a clean working tree;
replace the --dirty branch logic by computing changed files against the PR base
and passing those paths to Pint: fetch the base branch (origin/${{
github.base_ref }}), run git diff --name-only --diff-filter=ACMRTUXB origin/${{
github.base_ref }}...HEAD, filter results for files Pint cares about (e.g.,
*.php, *.blade.php, *.js, *.ts, *.vue), and if the list is non-empty invoke
vendor/bin/pint with those file paths (and tee output to pint_output.log),
otherwise skip/echo "no changed files to lint"; keep the PINT_MODE variable and
pint_output.log usage but replace the --dirty invocation with this git-diff
based path list logic.

In @.github/workflows/quickstart.yml:
- Around line 64-65: The workflow runs the "Run database migrations" step which
executes "php artisan migrate --force" but the MySQL service and DB environment
variables are commented out, causing migrations to fail; either remove or
conditionally skip the "Run database migrations" step when the DB service is not
present, or restore/uncomment the MySQL service and required DB env vars so the
"Run database migrations" step has a database to connect to—edit the job
containing the "Run database migrations" step to implement the chosen option
(skip/remove the step or uncomment the MySQL service and DB variables).
- Around line 59-62: The workflow step "Prepare Laravel environment" fails
because .env.testing.example is missing; either add a .env.testing.example at
repo root containing at minimum APP_KEY= (placeholder) and testing DB
credentials (DB_CONNECTION, DB_DATABASE, DB_USERNAME, DB_PASSWORD) and any
CI-specific test settings, or change the workflow's Prepare Laravel environment
step to use an existing template (e.g., copy .env.example) or generate the .env
from GitHub secrets/ENV vars before running php artisan key:generate; update the
workflow accordingly to ensure the cp command succeeds.

In @.github/workflows/README.md:
- Line 40: The README line claiming the release workflow "Automatically runs on
every push to the `master` branch" is incorrect; open the release workflow file
(release.yml) and update the README sentence to state the workflow is triggered
manually via workflow_dispatch (or mention it requires manual dispatch),
ensuring the documentation matches the actual trigger configured in release.yml.

- Around line 76-79: The directory existence check currently tests for
"resources/lang/en" (seen in the if conditional and the echo message) which is
wrong for CodeIgniter; update the conditional to check for the CodeIgniter
translations directory (e.g. "application/language/" or
"application/language/english") instead and adjust the error message to
reference the new path, keeping the same usage of LOG_FILE and the existing
tee/exit behavior.

- Around line 194-201: The current "Check for changes" step (id: check-changes)
only checks git diff for yarn.lock which can miss changes to package.json;
update the condition so it checks both yarn.lock and package.json (i.e., run git
diff --quiet against yarn.lock and package.json together) and set GITHUB_OUTPUT
changes_detected accordingly to ensure PRs are created when package.json changes
even if yarn.lock is unchanged.

- Line 177: The code uses mb_rtrim when constructing $fullPath which requires
PHP 8.4+; replace the mb_rtrim call with a compatible alternative (e.g.,
rtrim($baseDirectory, DIRECTORY_SEPARATOR)) or a small polyfill helper so
$baseDirectory is trimmed of DIRECTORY_SEPARATOR before concatenating
$basename['filename']; update the expression that builds $fullPath to use the
compatible trim function and keep the DIRECTORY_SEPARATOR concatenation logic
intact.
- Line 87: The code calls mb_rtrim() when constructing $realBaseWithSep inside
validate_file_in_directory(), which breaks on PHP <8.4; replace
mb_rtrim($realBase, DIRECTORY_SEPARATOR) with the single-byte safe
rtrim($realBase, DIRECTORY_SEPARATOR) to restore compatibility and preserve the
intended behavior of trimming the DIRECTORY_SEPARATOR before appending it.

In `@application/helpers/mailer_helper.php`:
- Around line 247-260: The referer validation using str_starts_with($referer,
$base_url) is vulnerable to host spoofing; instead parse both URLs and compare
scheme, host and port to ensure same origin before trusting it. In the block
that sets $referer/$base_url and before assigning $redirect (refer to
$_SERVER['HTTP_REFERER'], base_url(), $referer, $base_url, and
redirect($redirect)), use parse_url() on both $referer and $base_url, normalize
missing ports (80/443) and schemes, and only set $redirect = $referer when
parsed scheme, host and port exactly match; otherwise fall back to base_url().
Ensure empty/invalid referer parse results are rejected.

In `@application/language/english/ip_lang.php`:
- Around line 494-496: Update the language strings to capitalize "Croatian" and
make the label generic: change the value for '417_barcode_settings_enable' from
"Enable PDF 417 Barcode for croatian clients" to a generic label like "Enable
PDF 417 Barcode" (and update '417_barcode_settings_enable_hint' to remove the
country reference while keeping the instruction about recipient, IBAN and BIC).
Keep any Croatia-specific behavior in application logic (not in the language
key) so the UI label is reusable for other countries; adjust code that checks
country-specific rules to enforce display only for Croatian clients.
- Around line 494-505: The language keys starting with a digit (e.g.,
'417_barcode_settings', '417_barcode_settings_enable', etc.) violate the file's
naming convention; rename them to a text-prefixed form such as
'pdf_417_barcode_settings', 'pdf_417_barcode_settings_enable',
'pdf_417_barcode_settings_enable_hint',
'pdf_417_barcode_settings_recipient_street_name',
'pdf_417_barcode_settings_recipient_street_number',
'pdf_417_barcode_settings_recipient_city',
'pdf_417_barcode_settings_recipient_zip_code',
'pdf_417_barcode_settings_recipient_name',
'pdf_417_barcode_settings_recipient_iban',
'pdf_417_barcode_settings_recipient_bic',
'pdf_417_barcode_settings_remittance_text', and
'pdf_417_barcode_settings_remittance_text_tags' (or a consistent alternative
like 'pdf417_barcode_*') and update any code references that use the old keys to
the new names to keep lookups working.

- Around line 20-32: The tryFrom method incorrectly treats array_search index 0
as “not found” and returns the array index instead of the matched value; update
tryFrom to check search success with strict comparison against false (e.g.,
$searchResult !== false) and when found set $returnObject->value to the matched
value from getValues (e.g., $values[$searchResult]) rather than the index; also
ensure the method returns null (or nothing) when not found to preserve original
behavior.

PR #1241, opened 2025-03-09
Add pagination to invoice and quote templates

onny:
## Description

This will add pagination to the default invoice and quote pdf templates

image

By doing it this way, one could still modify the template to remove pagination or move it to a different place inside the document.

## Related Issue

Fixes https://github.com/InvoicePlane/InvoicePlane/issues/1234

## Pull Request Checklist

  * [ ] My code follows the code formatting guidelines.

  * [ ] Bugfix
  * [x] New Feature

### DIFF
--- a/application/views/invoice_templates/pdf/InvoicePlane - overdue.php
+++ b/application/views/invoice_templates/pdf/InvoicePlane - overdue.php
@@ -277,14 +277,20 @@
 
 <watermarktext content="<?php _trans('overdue'); ?>" alpha="0.3" />
 
-<footer>
+<div class="invoice-terms">
     <?php if ($invoice->invoice_terms) : ?>
         <div class="notes">
             <b><?php _trans('terms'); ?></b><br/>
             <?php echo nl2br(htmlsc($invoice->invoice_terms)); ?>
         </div>
     <?php endif; ?>
-</footer>
+</div>
+
+<htmlpagefooter name="footer">
+    <footer>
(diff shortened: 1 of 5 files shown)


application/views/invoice_templates/pdf/InvoicePlane - overdue.php
```diff
@@ -286,5 +286,11 @@
     <?php endif; ?>
 </footer>
 
+<htmlpagefooter name="footer">
```
nielsdrost7, line 289: Is that a new html tag?

onny: i have to look if this is a mpdf specific thing or we could use footer instead. the current footer part in the template is probably only text inserted at the end of the invoice and not a "real" footer shown on all pages?

nielsdrost7: > the current footer part in the template is probably only text inserted at the end of the invoice and not a "real" footer shown on all pages

correct, I think 🤔

onny: so the strange tag name comes from mpdf https://mpdf.github.io/reference/html-control-tags/htmlpagefooter.html
but inside the tag i'm using the footer tag to clarify the use case and put the invoice terms text into a div

nielsdrost7:
Thanks man!
Merged

PR #1401, opened 2026-01-20
Feat: new invoice status

nielsdrost7:


nielsdrost7 (commented):
@copilot

In application/modules/invoices/controllers/Invoices.php around lines 57-65 the
controller calls is_archived() but the method is missing in
application/modules/invoices/models/Mdl_invoices.php; add a public is_archived()
method to Mdl_invoices that applies the archived filter (e.g., adds the WHERE
condition used for archived invoices, such as invoice_is_archived = 1), returns
$this for method chaining consistent with the existing is_unpaid()/is_partial()
implementations, and include a brief doc comment and unit/integration check if
present in the model test suite.

application/modules/invoices/models/Mdl_invoices.php around lines 519-529: add a
new public method is_archived() matching the style of is_partial() and
is_unpaid() that applies the archived invoice status filter and returns $this;
implement it by calling $this->filter_where('invoice_status_id', <archived
status id or use the project constant/enum for archived status if available>)
and then returning $this so the controller can call is_archived() successfully.

In application/views/invoice_templates/pdf/InvoicePlane - partial.php around
lines 168 to 176, the condition incorrectly checks for 'Lead Source' before
rendering the Sales Person row; change the conditional to check for the 'Sales
Person' custom field (e.g. use isset($custom_fields['invoice']['Sales Person']))
so the Sales Person row only displays when that specific field exists, leaving
the rest of the markup unchanged.

In application/views/invoice_templates/pdf/InvoicePlane - partial.php around
lines 312 and 319, remove the hard-coded aroxmarketing.com policy/terms links or
replace them with a configurable or placeholder value; update the template to
either omit those links entirely, pull URLs from a settings variable (e.g.,
settings('company_policy_url') / settings('company_terms_url')), or insert
generic placeholder text like "{POLICY_URL}" and "{TERMS_URL}" so downstream
code or site admin can supply organization-specific links.

In application/views/invoice_templates/pdf/InvoicePlane - unpaid.php around
lines 222 to 230, the conditional that decides whether to render the "Sales
Person" row mistakenly checks isset($custom_fields['invoice']['Lead Source'])
instead of verifying the Sales Person field; change the condition to check for
the Sales Person key (e.g., isset($custom_fields['invoice']['Sales Person']) or
!empty(...) as appropriate) so the row only displays when the Sales Person
custom field actually exists, and keep the existing echo of
$custom_fields['invoice']['Sales Person'] as-is.

In application/views/invoice_templates/pdf/InvoicePlane - unpaid.php around line
228, the Sales Person custom field is output without escaping, creating an XSS
risk; update the output to use PHP's htmlspecialchars() with ENT_QUOTES and
UTF-8 (matching the other custom fields) so the value is safely escaped before
echoing.

In application/views/invoice_templates/pdf/InvoicePlane - unpaid.php around
lines 366 and 373, the footer contains hard-coded company-specific URLs to
"aroxmarketing.com"; remove these fixed links and replace them with configurable
placeholders or settings variables (e.g., policies_url and terms_url) or simply
remove the anchor tags so the template is generic; implement by reading two
optional config/settings values (with empty/null fallback so nothing prints if
not set) or substitute plain placeholder text like "{POLICIES_URL}" and
"{TERMS_URL}" that can be replaced by the app, ensuring no hard-coded domain
remains in the template.

In assets/core/scss/_core.scss around lines 580 to 583, the CSS selector has a
typo: change the class name from "archvied" to "archived" so the selector reads
&.archived, &.draft { background-color: lighten($color_status_draft, 10%); } to
ensure archived status labels apply correctly.

PR #1406, opened 2026-01-20
Feature: datatables sortable

nielsdrost7:


nielsdrost7 (commented):
@copilot this was a proof-of-concept for making index tables in InvoicePlane like datatables:

```
    $(document).ready(function() {
        $("#sort").DataTable({
            "paging": false,
            "searching": false,
            "info": false,
        });
    });
```

Now apply this same to all the index tables in InvoicePlane:

- [ ] clients
- [ ] invoices
- [ ] quotes
- [ ] products
- [ ] payments
- [ ] projects
- [ ] tasks
- [ ] users
- [ ] etcetera

keep it simple for now.
- [ ] Datatables,
- [ ] unique "id" per index table (to target)
- [ ] include the js in the Grunt script to compile (for now)
- [ ] Leave the js in the place where the original author included it in the client_table partial
- [ ] Don't run phpunit tests or anything

nielsdrost7 (commented):
@copilot
In `@application/modules/invoices/views/partial_invoice_table.php`:
- Line 2: The server-side dropup calculation causes incorrect dropdown
positioning after DataTables sorting; update the client to compute dropup
dynamically by attaching a show.bs.dropdown handler on the `#invoice-table` rows
(e.g., selector ".options") that measures the dropdown's offset and viewport
bottom and toggles the "dropup" class accordingly (or remove the server-side
dropup logic and rely solely on this JS); ensure this logic runs after
initializing the DataTable so it responds to row reordering and apply the same
change to the other table region referenced around lines 141-148.

In `@application/modules/layout/views/includes/head.php`:
- Line 13: The DataTables CSS loaded in includes/head.php uses CDN version
1.12.1 which mismatches the locally bundled JS (1.10.22) and creates an external
dependency; update includes/head.php to either point the <link> to the matching
CDN CSS version 1.10.22 or (preferred) bundle the CSS locally: add a copy task
for the DataTables CSS in Gruntfile.js, ensure the CSS is copied into the same
assets directory as other styles, and update includes/head.php to reference it
via _core_asset() like the other stylesheets (match package.json's DataTables
version and file name when adding the copy task and the _core_asset()
reference).

In `@application/modules/payments/views/partial_payments_table.php`:
- Around line 67-75: The DataTable initialization in the document-ready block
can error if the partial is loaded multiple times; update the init to first
check whether the table is already initialized using the DataTables guard (e.g.,
$.fn.dataTable.isDataTable or $.fn.DataTable.isDataTable) for the selector
"#payment-table" and only call $("#payment-table").DataTable(...) when it is not
already initialized (or alternatively call .DataTable().destroy() before
re-initializing if re-creation is desired); adjust the code inside the existing
$(document).ready(...) block so the guard prevents re-initialization.

In `@application/modules/quotes/views/partial_quote_table.php`:
- Around line 98-105: The dropup placement computed server-side (the $dropup
class) becomes incorrect after DataTables reorders rows; update the DataTable
initialization for "#quote-table" to either set "ordering": false to disable
sorting or attach a handler to the "draw.dt" event to recalculate and reapply
the $dropup placement for visible rows after each draw (recompute which rows
should have the dropup class and toggle it on the relevant row dropdown elements
inside the DataTable draw callback).

In `@application/modules/tax_rates/views/index.php`:
- Around line 69-74: The DataTable initialization for the tax-rate table
currently lets DataTables auto-sort the first column; update the
$("#tax-rate-table").DataTable({...}) options to include "order": [] so the
server-defined row order is preserved on load. Modify the DataTable init in the
tax rates view (the $("#tax-rate-table").DataTable call) to add the "order": []
option and replicate this pattern for all other DataTable initializations
(users, units, tasks, projects, quotes, products, payment_methods, payments,
invoice_groups, invoices, email_templates, families, clients) to ensure none
auto-sort on initialization.

In `@package.json`:
- Around line 23-24: Update the vulnerable datatables packages by bumping
"datatables.net" and "datatables.net-bs" to a patched release (minimum 1.10.23
for prototype-pollution fix, but prefer 1.11.3 or newer to cover XSS fixes);
change the versions for the package names "datatables.net" and
"datatables.net-bs" in package.json, then reinstall (npm/yarn), regenerate the
lockfile, run npm/yarn audit (or Snyk) to confirm no remaining advisories, and
run the app's UI/tests to ensure nothing broke after the upgrade.

@copilot apply changes based on the comments in [this thread](https://github.com/InvoicePlane/InvoicePlane/pull/1407#pullrequestreview-3680968846)

PR #1576, opened 2026-06-10
Fix 7 pre-beta bugs found in security-hardening review

nielsdrost7:
1. CRITICAL — Revert guest/View to Base_Controller; remove where_in guards that
   required login to view invoices/quotes by URL key, breaking all emailed links
   and payment-gateway success redirects.

2. HIGH — Fix create() in Mdl_Invoices/Mdl_Quotes bypassing the save() encryption
   override (parent::save vs $this->save dispatch). New records now encrypt
   invoice_password / quote_password on creation.

3. HIGH — Add try/catch to decrypt_password() in new Password_Encryption_Trait so
   that legacy plaintext passwords (stored before at-rest encryption landed) do not
   throw an uncaught Cryptor exception, crashing the admin view and PDF generation.
   Values migrate transparently on next save.

4. HIGH — Fix save() null-path: when called with no explicit db_array, populate
   from db_array() before encrypting so the password field is never written
   unencrypted via the no-arg path.

5. HIGH — Add get_by_id null-check to Quotes::delete(), matching the guard already
   present in Invoices::delete().

6. MEDIUM — Add xss_clean() to quote_password in quotes Ajax controller, matching
   the existing sanitization on invoice_password.

7. CLEANUP — Extract shared encrypt/decrypt logic to Password_Encryption_Trait,
   eliminating the verbatim duplication between Mdl_Invoices and Mdl_Quotes.

https://claude.ai/code/session_01WaPtptKZ2DgixiVmfyrWZK

### DIFF
--- a/application/core/Password_Encryption_Trait.php
+++ b/application/core/Password_Encryption_Trait.php
@@ -1,0 +1,63 @@
+<?php
+
+if ( ! defined('BASEPATH')) {
+    exit('No direct script access allowed');
+}
+
+/*
+ * InvoicePlane
+ *
+ * @author      InvoicePlane Developers & Contributors
+ * @copyright   Copyright (c) 2012 - 2018 InvoicePlane.com
+ * @license     https://invoiceplane.com/license.txt
+ * @link        https://invoiceplane.com
+ */
+
+/**
+ * Password Encryption Trait.
+ *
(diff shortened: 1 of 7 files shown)


nielsdrost7:
@coderabbitai full review

PR #1701, opened 2026-09-03
Fix DB_PASSWORD with special characters causing clueless 500 during setup

### LINKED ISSUE #1700: DB_PASSWORD with special characters (#, $, ;) causes clueless 500 during setup
### Title

If you happen to have '#' or '$' or ';' in your database password installer breaks with 500 error (no clear log)

### Expected Behavior

I'd expect the ipconfig.php.example to set a sample password within quotes like 'YOUR_PASSWORD'. Then automatically people would fill within those quotes (I presume). Else add a comment warning about these restrictions.

### Current Behavior

In my case the password had been DB_PASSWORD=#Ex4mpl3$Pass! - the installer broke with 500 error, leaving no trace in the server log. The internal log did write - however, was not helpful locating the issue.

### Steps to Reproduce

Upload fresh 1.7.2 and choose to use a password for your working database that contains invalid chars.
Even with debug=true the log may not give a non-dev the right clue.
In my case, a raw mysqli connection via CLI/PHP worked fine with the same credentials, which sent me digging elsewhere before I came back to check ipconfig.php parsing.

### Screenshots (Optional)

```textarea

```

### Possible Solution (Optional)

Wrap the sample password in single or double quotes in ipconfig.php.example, e.g. DB_PASSWORD='YOUR_PASSWORD', and/or add a comment above the line warning about #/$/other special characters needing quotes.

### Context (Environment)

_No response_

### Failure Logs (Optional)

_No response_

nielsdrost7:
# Pull Request Checklist

## Checklist
- [x] My code follows the code formatting guidelines.
- [x] I have tested my changes locally.
- [x] I selected the appropriate branch for this PR.
- [x] I have rebased my changes on top of the selected branch.
- [x] I included relevant documentation updates if necessary.
- [x] I have an accompanying issue ID for this pull request.

---

## Description

A database password containing special characters (`#`, `$`, `;`, etc.) broke the installer with an opaque 500 error and no useful trace in the log.

Root cause: `ipconfig.php` is parsed by `vlucas/phpdotenv`. `Setup::write_database_config()` was writing credentials wrapped in double quotes with `addcslashes()` escaping. Verified directly against the real phpdotenv library (not simulated): double-quoted values run a real escape-sequence parser, where an unescaped backslash followed by anything other than `" \ $ f n r t v` throws a hard parse error for the *entire* config file — not just the affected setting. That's exactly the class of failure reported: a password like `#Ex4mpl3$Pass!` written this way, or any password containing a raw backslash, could make the whole `ipconfig.php` fail to load with a message that gives no hint the password is the cause.

Single-quoted phpdotenv values, by contrast, do zero escape-sequence processing — every character is taken literally until the closing quote. The only character that can't be represented this way is a literal single quote itself.

### Changes

- `Setup::write_database_config()` now wraps `DB_HOSTNAME` / `DB_USERNAME` / `DB_PASSWORD` / `DB_DATABASE` in single quotes with **no escaping at all** — values are stored byte-for-byte as submitted.
- `validate_db_config_parameter()` rejects a literal single quote in the password with a new `password_contains_quote` error and a clear, actionable message, also surfaced proactively in the password field's help text (not only after a failed submit).
- Removed the unused `sanitize_db_config_value()` helper — it escaped single quotes with a backslash, which doesn't actually work either, since single-quoted phpdotenv values have no escape processing to interpret that backslash.
- `ipconfig.php.example` now documents both quoting conventions for anyone hand-editing the file: single quotes (recommended, zero escaping) and double quotes (supported, but requires `\\`/`\"` escaping and has a `${VAR}` interpolation gotcha), plus a note that an unquoted value truncates silently at the first space or `#`.

### Verification

Manually verified end-to-end against the actual `vlucas/phpdotenv` source (cloned and autoloaded standalone), including the exact password from the reported issue (`#Ex4mpl3$Pass!`) and a backslash-heavy password, confirming both round-trip correctly with this fix and that the previous double-quote approach would throw `InvalidFileException` for the backslash case. `php -l` clean on every changed file.

This PR intentionally does not include the accompanying PHPUnit test suite for this fix — those will follow separately.

---

## Related Issue(s)
Fixes #1700

---

## Motivation and Context
Users who choose a database password with common special characters (`#`, `$`, `;`, quotes) currently hit a 500 error during setup with no actionable log message, as described in #1700. This is a correctness/reliability fix for the setup wizard's database configuration step.

---

## Issue Type (Check one or more)
- [x] Bugfix

---

## Screenshots (If Applicable)
N/A — backend/config-parsing fix, no UI layout changes. The only user-visible change is the password field's help text and a new validation error message.

## Summary by CodeRabbit

- **Bug Fixes**
  - Database setup now clearly rejects passwords containing a single quote, preventing invalid configuration files.
  - Database credentials are stored with consistent quoting, while other supported special characters remain usable.

- **Documentation**
  - Updated setup guidance to explain password storage and the single-quote restriction.
  - Expanded configuration documentation with clearer rules for quoting, substitutions, escaping, and multi-line values.


smithdesign77:
Nice fix, and the ipconfig.php.example docs are really thorough.
hostname/username/database also get written unescaped in single quotes now, but only the password explicitly rejects a '. Guessing that's fine since those three are already restricted to a narrow charset elsewhere – just flagging it to double-check, while we are here.

nielsdrost7:
@smithdesign77 
> Guessing that's fine since those three are already restricted to a narrow charset elsewhere – just flagging it to double-check, while we are here.

You are absolutely correct about it and frankly restricting that single quote from the allowed characters list pissed me off.
I'm not seeing such a restriction in Laravel, for example.

Maybe we'll add the workaround that **if** you want a single quote in your password, then just put the password in double quotes.
I just don't want to hold back merging a solution for something as simple as that single quote, you know.

Anyone that has single quote in his password could have in his mind that he needs to wrap his password in double quotes.
Anyone that has single **and** double quotes is.. (i'm not saying it)

smithdesign77:
I just wanted to point at the full picture as it was: no restricting comment in the config, plus an error log that didn't shed light either – potentially leading to unnecessary support requests or "doesn't install" board posts, which isn't actually the case.
Fixing either one is enough, IMHO; we run into password policies to follow every day anyway – at the bank, mail provider, whatnot. That's okay — and by far easier than catching something like 'password was NULL' on the log level.

nielsdrost7:
Thanks man! I'll merge it

